PR #47, opened 2024-06-06
Networking v2

anovv:
- Introduced `IOLoop` class which asynchronously polls sockets and triggers handlers (`IOHandler`) on each event
- Everything is `zmq.PAIR` socket - bidirectional socket, can send and receive.
- `DataReader` + `DataWriter` implement `IOHandler`
- ACKing mechanism, ACK batching, retries + tests, makes sure messages are delivered over unreliable channels
- Low watermarks + buffering - ensure queues are FIFO + discard duplicates
- Introduced `Buffer` (bytearray), and `BufferQueues` for data buffering. Different buffering policies - buffer per message and periodic flush (fill buffer until it is full or until timeout - default behaviour). Large messages (multiple buffers per one) are not supported
- `BufferMemoryTracker` keeps track of (shared) memory size used by `DataReader` and `DataWriter` queues, blocks if above limit to inflict backpressure
- Backpressure via shared channel works
- Introduced `TransferHandler` (runs on `TransferActor`) which is responsible for passing messages back and forth between peer nodes in a cluster (via single bidirectional node<->node TCP connection): 1 Actor per node, 1 `IOLoop` per Actor/node, 2 `TransferHandler` instances per loop/actor/node - one for forward data (`Direction.SENDER`), one for backwards data (`Direction.RECEIVER`). `TransferActor` orchestration is WIP

  High level flow is: each operator actor (process) has a `DataReader` and `DataWriter` (source operators have only `DataWriter`, sink operators only `DataReader`). 
  2 scenarios:
  - If operators are on the same node, their `DataReader`/`DataWriter`(s) connect directly (each of them has a socket per channel which just connect 1-to-1)
  - If operators are on different nodes: each node runs a single TransferActor (which runs 2 TransferHandlers described above). `DataReader`/`DataWriter` connect to corresponding TransferHandlers on TransferActor, these 2 TransferHandlers pass (and receive acks) to 2 `TransferHandler`s on peer node which then pass data to corresponding operator's `DataReader`/`DataWriter` on another node.

- [TODO] Credit-based Flow Control for backpressure
- [TODO] Networking metrics

All the necessary logic is in volga.streaming.runtime.network and some orchestration in volga.streaming.runtime.master.worker_lifecycle_controller.py and volga.streaming.runtime.worker.task.stream_task.py

volga.streaming.runtime.network.experimental and volga.streaming.runtime.network.deprecated can be skipped.


anovv:
Just realised that having 2 TransferHandlers per node means we have 2 tcp channels per node<->node connection, this should be fixed.

volga/streaming/runtime/network/transfer/local/data_reader.py
```diff
@@ -0,0 +1,132 @@
...
+        payload = get_payload(buffer)
+
+        # TODO implement partial messages
+        res = []
+        for (msg_id, data) in payload:
+            msg = simplejson.loads(bytes_to_str(data))
+            res.append(msg)
+        return res
+
+    def send(self, socket: zmq.Socket):
```
0xaa4eb, line 62: Seems to be not so obvious that `reader` has `send` method. When I looked into the method, it appears to send only `acks`, so maybe we can do something about it

volga/streaming/runtime/network/buffer/buffer_queues.py
```diff
@@ -0,0 +1,95 @@
...
+        self._buffer_pool = buffer_pool
+        self._buffer_queues: Dict[str, deque] = {c.channel_id: deque() for c in channels}
+        self._locks = {c.channel_id: threading.Lock() for c in channels}
+        self._last_update_ts = {c.channel_id: None for c in channels}
+        self._buffering_policy = buffering_policy
+
+    def append_msg(self, message: ChannelMessage, channel_id: str) -> bool: # backpressure or not
+        msg_str = simplejson.dumps(message)
+        lock = self._locks[channel_id]
+        lock.acquire()
```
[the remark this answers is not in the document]
anovv: Good catch!

volga/streaming/runtime/network/buffer/buffer_pool.py
```diff
@@ -0,0 +1,25 @@
...
+    def instance(node_id: str, capacity_bytes: int = DEFAULT_CAPACITY):
+        if BufferPool._instance is None:
+            BufferPool._instance = BufferPool(node_id, capacity_bytes)
+        return BufferPool._instance
+
+    def __init__(self, node_id: str, capacity_bytes: int = DEFAULT_CAPACITY):
+        self._capacity_bytes = capacity_bytes
+        self._node_id = node_id
+
+    def try_acquire(self, amount: int) -> bool:
```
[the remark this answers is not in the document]
anovv: What's called a "buffer_pool" here is not really a pool, we simply track size of used memory here. Renamed it to BufferMemoryTracker

volga/streaming/runtime/network/transfer/local/data_reader.py
anovv: This is because `DataReader` (as well as `DataWriter`) both implement `IOHandler` interface which has both `send() `and `rcv().` Not sure how properly name here to avoid confusion

pyproject.toml
```diff
@@ -12,6 +12,9 @@ ray = "2.22.0"
 pydantic = "^1.10.13"
 simplejson = "^3.19.2"
 pyzmq = "23.2.0"
+shared-memory-dict = "0.7.2"
+lockit = "1.0.1"
+aenum = "3.1.15"
```
0xaa4eb, line 17: new dep `locket` here

volga/streaming/runtime/network/buffer/buffer_memory_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, node_id: str, capacity_per_channel: int, capacity_per_out: int):
+        self._capacity_per_in_channel = capacity_per_channel
+        self._capacity_per_out = capacity_per_out
+        self._node_id = node_id
+        # TODO should we add job_name to dict and lock identifiers?
+        self._shared_dict_name = f'shared-dict-{self._node_id}'
+
+        os.makedirs(BufferMemoryTracker.LOCKET_LOCK_DIR, exist_ok=True)
+        self._lock_name = f'{BufferMemoryTracker.LOCKET_LOCK_DIR}/lock-memory-allocator-{self._node_id}'
```
0xaa4eb, line 43: let's use unique folder for tests

volga/streaming/runtime/network/buffer/buffer_memory_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+        if node_id not in BufferMemoryTracker._instances:
+            BufferMemoryTracker._instances[node_id] = BufferMemoryTracker(node_id, capacity_per_in_channel, capacity_per_out)
+        return BufferMemoryTracker._instances[node_id]
+
+    def __init__(self, node_id: str, capacity_per_channel: int, capacity_per_out: int):
+        self._capacity_per_in_channel = capacity_per_channel
+        self._capacity_per_out = capacity_per_out
+        self._node_id = node_id
+        # TODO should we add job_name to dict and lock identifiers?
+        self._shared_dict_name = f'shared-dict-{self._node_id}'
```
0xaa4eb, line 40: let's use unique name for tests

volga/streaming/runtime/network/buffer/buffer_memory_tracker.py
anovv: Passed `job_name` to make lock name unique

volga/streaming/runtime/network/buffer/buffer_memory_tracker.py
anovv: Passed `job_name` to make memory block name unique

pyproject.toml
anovv: Done

PR #139, opened 2026-05-30
use bitmask-oriented id abstractions

AkhtemVeis:
The point of the PR is:
1. Replace string-based identifiers with typed numeric IDs on the hot path.
2. Give each ID-like concept its own struct/enum so the type system enforces correctness.
3. Reduce allocations and improve cache efficiency.

OperatorTypeCode — a #[repr(u8)] enum that assigns a stable numeric code to each concrete operator variant. Restricts the system to at most 255 distinct operator types. Codes must only be appended (never renumbered) so existing checkpoints and serialized IDs stay interpretable.
OperatorId — a u16 identifier for a logical operator node within the execution graph. Layout: [op_type: u8 | instance: u8], where the upper 8 bits hold the OperatorTypeCode and the lower 8 bits hold the instance index of that operator type within the graph. All parallel partitions of the same operator share one OperatorId.
VertexId — a u32 identifier for a single execution vertex (one parallel partition of an operator). Layout: [op_type: u8 | instance: u8 | task_index: u16], where the upper 16 bits form the OperatorId and the lower 16 bits identify the parallel partition within that operator.
ChannelId — a u64 globally unique identifier for a directed channel between two vertices. Layout: [source: VertexId | target: VertexId], upper 32 bits for the source and lower 32 bits for the target.
Also added: Display/Debug formatters, accessor methods, and unit tests covering round-trips, layout packing, and display output.

Also added formatters, helper methods and tests.

### DIFF
--- a/src/common/ids.rs
+++ b/src/common/ids.rs
@@ -331,32 +331,13 @@ mod tests {
     #[test]
     fn op_type_code_roundtrip_covers_all_variants() {
         // Exhaustively iterate via from_u8 across the assigned range.
-        let all = [
-            OperatorTypeCode::Invalid,
-            OperatorTypeCode::SourceVector,
-            OperatorTypeCode::SourceWordCount,
-            OperatorTypeCode::SourceDatagen,
-            OperatorTypeCode::SourceHttpRequest,
-            OperatorTypeCode::SourceKafka,
-            OperatorTypeCode::SourceParquet,
-            OperatorTypeCode::SinkInMemoryStorageGrpc,
-            OperatorTypeCode::SinkRequest,
-            OperatorTypeCode::SinkParquet,
-            OperatorTypeCode::Map,
-            OperatorTypeCode::KeyBy,
-            OperatorTypeCode::Reduce,
-            OperatorTypeCode::Aggregate,
-            OperatorTypeCode::Window,
-            OperatorTypeCode::WindowRequest,
-            OperatorTypeCode::Join,
-            OperatorTypeCode::Chained,
-        ];
-
+        let all: Vec<_> = (0u8..=u8::MAX)
+            .filter_map(OperatorTypeCode::from_u8)
+            .collect();
         for code in all {
             let byte = code.as_u8();
             assert_eq!(OperatorTypeCode::from_u8(byte), Some(code), "code {byte:#x}");
         }
-
         // Unassigned bytes return None.
         assert_eq!(OperatorTypeCode::from_u8(0x12), None);
         assert_eq!(OperatorTypeCode::from_u8(0xFF), None);
--- a/src/common/ids.rs
+++ b/src/common/ids.rs
@@ -9,6 +9,7 @@
 //! - up to 255 instances of the same operator type within one pipeline
 //! - up to 65535 parallel partitions per operator instance
 
+use num_enum::TryFromPrimitive;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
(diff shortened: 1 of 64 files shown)


src/common/ids.rs
```diff
@@ -0,0 +1,459 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn op_type_code_roundtrip_covers_all_variants() {
+        // Exhaustively iterate via from_u8 across the assigned range.
+        let all = [
```
anovv: Can we somehow iterate OperatorTypeCode values itself instead of copying into separate all (ie if we add new OperatorTypeCode we may forget to update the test) ? Is it possible to iterate enum values in Rust?

src/common/ids.rs
```diff
@@ -0,0 +1,459 @@
...
+            0x0B => Some(Self::KeyBy),
+            0x0C => Some(Self::Reduce),
+            0x0D => Some(Self::Aggregate),
+            0x0E => Some(Self::Window),
+            0x0F => Some(Self::WindowRequest),
+            0x10 => Some(Self::Join),
+            0x11 => Some(Self::Chained),
+            _ => None,
+        }
+    }
```
anovv, line 67: My concern is the hard binding of the type to the byte—adding/removing operator types can lead to confusion: it'll be necessary to update not only the enum but also the from_u8 function. Also right now, everything is in order (first sources, then regular, then sinks) which looks good and pretty, but what if we add another source? It'll go down after the sink - looks awkward. Functionally it's ok, just curious if we can do better

anovv, line 67: Asked Gemini, suggested looking at num-derive and num-traits crates and strum crate (this one can turn enums to iterators, can be relevant) - may be worth exploring. In general im fine with the way it is now, just if it can be improved with not too much work then great, if it's too much work - dont bother, keep as is.

src/runtime/observability/metrics/mod.rs
```diff
@@ -264,23 +264,30 @@ impl OperatorAggregateMetrics {
...
+        let mut metrics_by_operator: HashMap<OperatorId, Vec<TaskMetrics>> = HashMap::new();
 
-        // Group task metrics by operator_id
-        for (vertex_id, task_metrics) in tasks_metrics.iter() {
-            if let Some(vertex) = graph.get_vertex(vertex_id) {
-                let operator_id = vertex.operator_id.clone();
+        // Group task metrics by operator_id.
+        // The HashMap is keyed by vertex_id rendered via Display; iterate vertices in the graph
+        // and reverse-match by Display form to recover their operator_id.
+        let by_label: HashMap<VertexId, &TaskMetrics> = tasks_metrics
```
anovv: Can you explain why we need separate by_label map here?

src/runtime/tests/checkpoint_recovery_test.rs
```diff
@@ -3,6 +3,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use datafusion::common::ScalarValue;
+use kameo::Reply;
```
anovv: is this needed?

anovv (approved):
Thanks for working on this, makes things more consistent and structured. 

Looks good to me, just few questions/suggestions

src/common/ids.rs
AkhtemVeis: Yeah, we can figure it out.

src/common/ids.rs
AkhtemVeis: I researched the possibilities and came up with the following alternatives:

To get rid of additional efforts for from_u8 update I found the cleanest way with new crate "num_enum", which is just a codegen library, which will do basically the same thing in the code when calling it's method.
The code would look like this:
```
use num_enum::TryFromPrimitive;
#[repr(u8)]
#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug, Serialize, Deserialize, TryFromPrimitive)]
pub enum OperatorTypeCode {
    Invalid        = 0x00,
    SourceVector   = 0x01,
    // ...
    Chained        = 0x11,
}
impl OperatorTypeCode {
    pub fn from_u8(byte: u8) -> Option<Self> {
        Self::try_from(byte).ok()
    }
    // ...
}
```

For ordering types I think the only way we have is introducing strict intervals for source, regular and sinks.
My proposal is to reserve 32 values for both source and sinks and the rest for regular transforms. What do you think?

AkhtemVeis: I think I can fix it quickly

src/runtime/observability/metrics/mod.rs
AkhtemVeis: Totally agree there is no need for it - looks like Claude overthought it...
Refactored this part.

src/runtime/tests/test_utils.rs
```diff
@@ -54,8 +49,8 @@ pub fn window_rows_from_messages(messages: Vec<crate::common::message::Message>)
 
         let upstream_vertex_id = msg.upstream_vertex_id();
         let upstream_task_index = upstream_vertex_id
-            .as_deref()
-            .and_then(parse_task_index_from_vertex_id);
+            .unwrap()
+            .task_index();
```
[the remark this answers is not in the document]
AkhtemVeis: e925ae1430bc72230747042137c5e057

src/common/ids.rs
AkhtemVeis: Implemented it.

AkhtemVeis: Fixed

src/common/ids.rs
```diff
@@ -0,0 +1,430 @@
...
+            Self::Chained => "Chained",
+        }
+    }
+}
+
+impl fmt::Display for OperatorTypeCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.short_name())
+    }
+}
```
anovv, line 97: we can also simplify `short_name()` and custom `Display` impl using `strum` crate - smth like

```
use strum::{Display, IntoStaticStr};

#[derive(Display, IntoStaticStr)] // implements Display and &'static str conversion
pub enum OperatorTypeCode {
...

```
then short name is simply


```
pub fn short_name(self) -> &'static str {
    self.into() // refers to IntoStaticStr
}

```
(the syntax may not be exact - it is AI generated, but you get the point)

and custom Display won't be needed either (unless we want to use short_name() directly).

PR #154, opened 2026-07-21
Redesign window state store and RANGE evaluation

anovv:
## Summary
- replace the SortedKV/BucketIndex-specific window layer with explicit `WindowOperatorStore` and coherent `WindowRequestStore` contracts
- split WO advancement into emit discovery, exact coverage planning, selective historical reads, and slide/rebuild evaluation
- keep WRO rebuild-only with one coherent merged load across requested points and windows
- simplify module boundaries, centralize shared models and accumulator helpers, and flatten evaluation/tile code
- provide a standalone window architecture README and focused high-signal tests

## Test plan
- [x] `cargo test runtime::operators::window` (58 passed)
- [x] `git diff --check`


src/runtime/operators/window/aggregates/mod.rs
```diff
@@ -23,66 +17,24 @@ use crate::runtime::operators::window::top::accumulators::ratio::{
...
-        panic!("Aggregation must implement produce_aggregates_from_ranges")
-    }
-}
-
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AggregatorType {
-    PlainAccumulator, // incremental updates, runs on whole window, supports tiling
-    RetractableAccumulator, // incremental updates and retracts, only updates state with new events
-    Evaluator, // evaluates whole window
+    PlainAccumulator,
```
AkhtemVeis: I think we might need comments in the future:)

Issue #156, opened 2026-07-28
Make execution_attempt_id unqiue across master restarts


anovv:
execution_attempt_id is used for zombie writers fencing, but currently is generated based on local master state (it is a simple counter). We should make sure if master dies new execution_attempt_id are unique and not repeating. Also do they need to be incremental?

anovv:
This is now a hard blocker for the Scylla window store ([#157](https://github.com/volga-project/volga/pull/157), umbrella [#291](https://github.com/volga-project/volga/issues/291)), and the answer to "do they need to be incremental?" is yes.

The store stamps every row with `Version = (attempt, epoch)` and filters reads against a per-key-group history of `(attempt, epoch)` prefixes. The order has to be total and the ordering has to be *durable*, because Scylla data outlives the master process. Two things follow:

- **Monotonic, not just unique.** A UUID would fence zombies but would not order two attempts, so the store could not tell a pre-crash write to keep from a post-checkpoint write to exclude.
- **Derived from durable state, not local counters.** `attempt_id` resets to 0 on pipeline start, and a new incarnation at attempt 0 sits below every existing cut — its writes get filtered out and its cells lost to last-write-wins. `max(restored cut attempts) + 1` is also not enough: attempts 2 and 3 commonly restore from the same checkpoint, so both compute the same value and share a version space with each other's zombie.

Requirement: the master allocates from durable state (`max(observed) + 1`), persists it **before** configuring workers, and passes it to tasks with restore data. Workers assert the new attempt exceeds every attempt in any restored cut history, and the backend refuses to open until this exists.

anovv:
Landed in #307. The master allocates `last+1` (or `0`) from the checkpoint store and persists it before configure, so recoveries from the same checkpoint no longer share a version space. Recovery budget is counted separately from the id.

What is still in-memory is the checkpoint store itself — a master *process* death still loses both completed checkpoints and the last attempt. That is #158, not more work on this issue.

PR #157, opened 2026-07-29
Scylla window store design: versioned cut history with checkpoint publication

anovv:
Normative design for using Scylla as the remote state store behind the window
operator. This is the plan of record; it replaces the protocols drafted in
#298 and #299, both now closed.

Two ideas carry the whole thing.

**1. `Version = (attempt, epoch)`, visibility is a per-key-group cut
history.** Every row is stamped with a totally ordered version, and what a
reader sees is a short sorted list of at most one `(attempt, epoch)` prefix
per attempt. Older attempts keep their prefix forever, so pre-crash history
never disappears; a zombie's post-checkpoint writes sit above its prefix and
are excluded permanently, by every later attempt. `epoch` is a per-key-group
write-batch counter that starts at zero each attempt and needs no seeding,
because `attempt` is the high-order component of the order.

**2. Publication is the checkpoint.** On checkpoint *completion* the WO
writes its cut, the committed watermark and a retention floor into one
`window_kg_meta` row per key group. That single row is the entire interface
between the streaming side and the serving side.

The second idea is what makes the design small. Because the published cut is
exactly the cut a successor restores from, it only ever grows — so there is
no publish cadence, no ownership lease, no per-key pin, no `prev_E`, and no
clamp. The clamp was the largest source of complexity in the earlier drafts,
and it exists only if a reader can be shown writes belonging to a checkpoint
that later aborts. Waiting for completion removes the possibility instead of
repairing it.

### Read contract

Reads say what they want to see, which is the seam where the WO emit path and
WRO could previously disagree about the same query:

- `Committed` — as of the last completed checkpoint. A window the WO emitted
  over `[lo, hi]` and a WRO `Committed` read of `[lo, hi]` return the same
  rows. Deterministic and replayable.
- `Fresh` — head-scoped best effort. Interior tiles are read at the cut; only
  raw above `committed_wm` is read at the current attempt, so the extra cost
  is one bounded raw slice. Two consequences are stated rather than
  engineered away: a torn commit can appear as a subset of one batch's events
  (never a corrupt aggregate, since Fresh widens raw only), and a late arrival
  below the head is invisible until its checkpoint completes.

WRO is not watermark-driven and does not need to be. It needs one thing the
watermark used to be a proxy for: whether the requested range still exists.
The published `retention_floor` gives it that directly, as a coverage guard
that refuses uncovered historical ranges. A normal "as of now" query always
passes it.

Retention itself is derived from the committed watermark and published, and
**GC deletes only below the published floor**. One rule, two properties: the
reader's coverage guarantee, and a floor that cannot move backwards across a
failover.

### Freshness, honestly

`Committed` read-your-writes across the WO/WRO split is checkpoint-granular.
For event-time correctness that is rarely the binding constraint — watermark
lag usually dominates publication lag — and `Fresh` covers the cases where it
is. Both are in *Accepted divergences* rather than buried.

### Two master prerequisites

Neither is window work, and the backend refuses to open without them.

- **Durable attempt allocation** (#156). `attempt_id` resets to `0` on
  pipeline start, and `max(restored) + 1` collides because attempts 2 and 3
  commonly restore from the same checkpoint.
- **A checkpoint-completion notification to tasks.** The master already
  tracks `latest_complete_checkpoint` and emits
  `LifecycleEvent::CheckpointCompleted`, but never tells the tasks. Cheaper
  fallback: piggyback the last completed id on the next barrier and publish
  one checkpoint behind.

### Also here

`docs/window-store-impl.md` carries the work items. The existing Scylla stack
(#287–#293) was written against the previous protocol and needs rework rather
than extension — `window_head`, `window_recovery_bases`, the lease, steal,
promote and `serving_publish` all go away. #291 is the umbrella tracking that.

### DIFF
--- a/src/runtime/operators/window/STORE_DESIGN.md
+++ b/src/runtime/operators/window/STORE_DESIGN.md
@@ -154,15 +154,16 @@ for the attempt:
 
 ```text
 StateNamespace  = (pipeline, owner operator)
+max_parallelism = job-wide, bound on the client
 KeyGroupRange   = groups for this task_index at (p, max_p)
 WriterId        = this task execution
 AttemptToken    = job-level execution attempt
 ```
 
-`stream_due` / `checkpoint` / `restore` use that bound range. They do not take
-`owned` or `namespace` — the client is the task's store. Per-key calls derive
-`key_group` from the business key and must land in the bound range (routing
-bug otherwise).
+The client derives `key_group` from `Key.hash` and bound `max_p`. It never
+trusts a caller-supplied group. `stream_due` / `checkpoint` / `restore` use
+the bound range and do not take `owned` or `namespace`. Per-key calls must
+land in the bound range (routing bug otherwise).
 
 WRO reuses the WO owner namespace. It addresses rows by `PartitionKey`
 (namespace + business key). Request routing still sends a lookup to the WRO
@@ -239,7 +240,9 @@ pub trait WindowRequestStore: Send + Sync + Debug {
 
 - Missing partitions and empty runs return empty/default results.
 - Raw rows are restricted to the requested runs, deduplicated, and globally
-  ordered by `Cursor`. Their Arrow schema includes `__seq_no: UInt64`.
+  ordered by `Cursor`. Their Arrow schema includes `__seq_no: UInt64`. The
+  Scylla backend stores **one CQL row per event**, clustered by time; it does
+  not pack events into a serialized batch blob.
 - Tiles are unique by `(granularity, tile_start)` and restricted to requested
   runs. Missing tiles mean empty intervals.
 - `commit_events` atomically publishes raw rows, replacement tiles, key state,
@@ -270,6 +273,13 @@ state and triggers for that range. The shared backend implements
 `OperatorStore::maintain` and prunes eagerly under the same retention contract
 as Scylla (control flow identical; lag ~0), reading the range from
 `OperatorTaskState` rather than a task-suffixed namespace blob.
+
+**Land the engine contract and InMem before any Scylla backend.** Today
+`StateNamespace` still includes `task_index`, and restore/maintain key on that
+blob. Operator-scoped namespace plus range isolation must ship on the traits
+and InMem first (two tasks, one worker: restore/maintain of task 0 must not
+clobber task 1). Otherwise the same type has two meanings and shared-worker
+restore wipes sibling keys. Scylla implements that already-landed contract.
 
 ## Runtime flows
 
@@ -404,6 +414,28 @@ watermark remain in the operator's `WindowStateSnapshot` envelope.
 
 ### Tables
 
+`attempt` is **not** in any data partition key. It is clustering (or a
+column) so recovery overlay is a visibility filter on one `LOCAL_QUORUM`
+read, not a second partition hop. Restore still does not copy rows.
+`maintain` drops unreachable generations so the filter stays one or two
+attempt clauses.
+
+**GC default: skinny index + per-key data PK. Do not cluster `business_key`
+under `(namespace, key_group, bucket)`.** `load_raw` is per-key; a key group
+is unbounded, so that clustering would make one partition (compaction,
+repair, hot token) hold every key in the group+bucket. Keep `business_key`
+in the data partition key so each key+bucket stays small.
+
+That data layout cannot list “this key group” or `bucket_start < data_floor`,
+so `commit_events` also writes a payload-free index
+`(namespace, key_group, bucket_start) → business_key`. `maintain` walks
+event-time `bucket_start` values below `data_floor` (configured bucket width;
+empty index partitions are skipped), pages each partition, and deletes the
+per-key data partitions. If the key count per group+bucket grows large, shard
+the index (`index_shard = hash(business_key) % N`) so each index partition
+stays bounded. TWCS/TTL on write-time cannot replace this: event-time buckets
+≠ write time, and replay rewrites old buckets.
+
 ```sql
 CREATE TABLE window_head (
     namespace blob,
@@ -428,32 +460,43 @@ CREATE TABLE window_recovery_bases (
     PRIMARY KEY ((namespace, recovery_attempt), range_start)
 );
 
+-- Skinny GC index only: no payloads. v1: one partition per group+bucket.
+-- If that partition grows too wide, add index_shard to the partition key
+-- (hash(business_key) % N) and page all shards in maintain.
+CREATE TABLE window_kg_buckets (
+    namespace blob,
+    key_group int,
+    bucket_start bigint,
+    business_key blob,
+    PRIMARY KEY ((namespace, key_group, bucket_start), business_key)
+);
+
 CREATE TABLE window_raw (
     namespace blob,
     key_group int,
     business_key blob,
-    attempt blob,
     bucket_start bigint,
     event_ts bigint,
     seq_no bigint,
+    attempt blob,
     epoch bigint,
     payload blob,
     PRIMARY KEY (
-        (namespace, key_group, business_key, attempt, bucket_start),
-        event_ts, seq_no, epoch
+        (namespace, key_group, business_key, bucket_start),
+        event_ts, seq_no, attempt, epoch
     )
 ) WITH CLUSTERING ORDER BY (
-    event_ts ASC, seq_no ASC, epoch DESC
+    event_ts ASC, seq_no ASC, attempt ASC, epoch DESC
 );
 
 CREATE TABLE window_tiles (
     namespace blob,
     key_group int,
     business_key blob,
-    attempt blob,
     granularity_ms bigint,
     bucket_start bigint,
     tile_start bigint,
+    attempt blob,
     epoch bigint,
     payload blob,
     PRIMARY KEY (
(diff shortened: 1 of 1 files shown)


src/runtime/operators/window/STORE_DESIGN.md
```diff
@@ -0,0 +1,594 @@
...
+
+Foyer is WO-only:
+
+```text
+meta:
+    PartitionKey -> (writer version, KeyState)
+data:
+    (PartitionKey, family, bucket) -> materialized writer-view data
+triggers:
+    (namespace, attempt, bucket, shard) -> immutable due entries
```
anovv: Do we need attempt here as a key?

src/runtime/operators/window/STORE_DESIGN.md
```diff
@@ -0,0 +1,594 @@
...
+### Recovery
+
+1. The replacement worker has the restored version in
+   `WindowBackendSnapshot::Versioned`.
+2. During operator restore, the worker's Scylla store creates the new attempt,
+   inserts `window_recovery_bases(new attempt -> restored checkpoint version)`,
+   and starts its epoch at zero. The row must exist before the attempt
+   publishes data. The master only plans and sends the restore snapshot.
+3. Source restores its checkpoint offset and replays post-checkpoint input.
+4. On first access to a key, retain its old serving version, claim ownership,
```
anovv: Why serving version, not writer version?

anovv:
We should also think on consistency guarantees per query - LOCAL, ONE, QUORUM etc., those affect latency

anovv:
Can we also use USING TIMESTAMP clasue with version/attempt to fence zombie writes without LWT?

src/runtime/operators/window/STORE_DESIGN.md
```diff
@@ -0,0 +1,607 @@
...
+  `trigger_shard` bound each Scylla partition:
+  `stable_hash(business_key) % trigger_shard_count`. Without it, every trigger
+  in one namespace, attempt, and time bucket would share one hot, potentially
+  oversized partition. `stream_due` queries all shards for each relevant time
+  bucket and merges their results by trigger cursor. The shard is only a
+  physical storage concern, not a window, task, or rescaling key group; its
+  count must remain stable for the namespace. Row triggers use a sentinel
+  `window_id`; scheduled windows retain their actual `WindowId`.
+
+Raw payloads are Arrow-IPC one-row `RecordBatch` values including `__seq_no`..
```
anovv: why one-row?

anovv:
we support keygroups now for key routing - redesign for this

anovv:
**Review: address before Scylla impl**

1. **GC / `maintain` vs PKs.** `(namespace, key_group, business_key, …)` as the partition key makes it impossible to list or drop “this key group” or `bucket_start < data_floor`. TWCS/TTL do not fix that (write-time ≠ event-time; replay rewrites old buckets). Put `business_key` in clustering, or add an explicit key/bucket index `commit_events` updates.

2. **`window_triggers` clustering ≠ `stream_due`.** `key_group` then `fire_ts` cannot slice a time range over a group range without `ALLOW FILTERING`. Cluster by `fire_ts` first; locality is `(bucket_start, kg_shard)`; filter `key_group` in the client.

3. **Recovery overlay must not be extra partition hops.** Keep `attempt` out of the partition key (clustering/columns only). Overlay is a visibility filter on one `LOCAL_QUORUM` read. Restore still does not copy rows. `maintain` drops unreachable generations so the filter stays 1–2 clauses.

4. **Engine contract + InMem first.** Today `StateNamespace` includes `task_index` and restore/maintain key on ns. Operator-scoped ns + range isolation must land on the traits and InMem (two tasks, one worker, no clobber) before Scylla. Otherwise the same type has two meanings and shared-worker restore wipes sibling keys.

5. **Also:** don’t GC a serving version still pinned by in-flight WRO; bind `max_p` on the per-task client and derive `key_group` (never trust the caller). `window_raw` stays one row per event, clustered by time; `commit_events` uses a same-partition `UNLOGGED BATCH` (one mutation per key+bucket, not one RTT per event). No logged/cross-partition BATCH. Do not pack events into serialized blobs.

anovv:
**GC index: `bucket_start` must be clustering, not a partition-key component**

`window_kg_buckets` is currently `PRIMARY KEY ((namespace, key_group, bucket_start), business_key)`. You cannot `SELECT … AND bucket_start < data_floor` — maintain would have to probe historical bucket ids (unbounded). Do **not** add a second `live_buckets` table.

**One index table:**

```sql
CREATE TABLE window_kg_buckets (
    namespace blob,
    key_group int,
    bucket_start bigint,
    business_key blob,
    PRIMARY KEY ((namespace, key_group), bucket_start, business_key)
) WITH CLUSTERING ORDER BY (bucket_start ASC, business_key ASC);
```

`commit_events`: one unversioned upsert (no `attempt`/`epoch`; do not say this is written “under `(job attempt, E)`”). Failed head LWT may leave a stale index row; `maintain` deletes it.

`maintain`, per owned `key_group`:

```cql
SELECT * FROM window_kg_buckets
 WHERE namespace = ? AND key_group = ?
   AND bucket_start < ?
```

Then delete those per-key raw/tile partitions and the matching index rows. After prune, the partition holds only live buckets: `keys_in_group × (retention / bucket_width)`, skinny cells. Shard with `index_shard` on the PK later if a hot group is too wide — not v1.

Update the GC default paragraph, the SQL, the `window_kg_buckets` bullet, WO write step 4, and the maintain body. Leave per-key data PKs / `window_raw` / triggers as they are.

anovv:
**Nits (non-blocking except the floor bind)**

1. **`SELECT` floor bind.** `bucket_start < data_floor` is wrong if `data_floor` is a timestamp and the bucket still covers `[data_floor, data_floor + width)`. Only drop buckets whose **entire** range is below the floor: bind `floor_bucket` so `bucket_start + width <= data_floor` (i.e. `bucket_start < floor_bucket`).

2. **MVCC generation GC key set.** You cannot scan `window_head` (per-key PK). The key set is keys still present in **live** `window_kg_buckets` rows for the owned groups.

3. **Foyer trigger cache** is still `(namespace, attempt, bucket, kg_shard)`. Trigger partitions are `(namespace, bucket, kg_shard)` — drop `attempt` from the cache key.

4. **Tile delete.** Index is per event-time bucket, not per granularity. For each index hit, drop every `granularity_ms` tile partition for that key+bucket (granularities from task state), not only `window_raw`.

anovv:
**Protocol update** (this PR is now the source of truth).

Streaming and request share versioned tables. Lease/publish are request-only.

- **WO overlay:** `attempt == me OR (cp.attempt AND E ≤ cp_E)`. Restore writer from CP. `E` never resets. Triggers use the same overlay.
- **Request lease** (`window_kg_lease`, one row per `key_group`): steal copies `prev ← serving`; `publish()` after `serving_wm`; WRO pins once, then `E ≤ serving_E AND NOT (attempt == prev.attempt AND E > prev.E)`, then latest per cell.
- **Removed:** per-key `window_head`, `window_recovery_bases`, HeadClaim/DashMaps, ingest LWT.

Stacked impl: `docs/window-store-impl.md` on this branch. **#288 must be rewritten** against this; do not land HeadClaim / recovery_bases / per-key steal-vs-promote as currently written.

anovv:
**Protocol tweak** after review — two spec holes, now closed in this PR (`78ce5a67`).

**(1) WO overlay was `me | cp.attempt`.** A single `StateVersion` is not a chain, so after a second failover idle keys left on an older attempt disappeared from WO (WRO still saw them via `E`). Overlay is now:

```text
attempt == me OR E <= cp_E    -- any attempt
```

**(2) `prev ← serving` is the last *published* cut, not the last writer.** If B steals, writes, and C steals before B publishes, B’s extras are not `prev.attempt`. Filter is now:

```text
E <= serving_E
AND (prev_E unset OR E <= prev_E OR attempt == serving.attempt)
```

Steal sets `prev_E ← serving_E` (`0` if serving empty). Publish does not touch `prev_E`. Dropped `prev_attempt`.

**(3)/(4)** follow from (2): extras above `prev_E` are only legal for `serving.attempt`; first steal with empty serving uses `prev_E = 0`.

**(5)** `serving_wm` is event-time only, not per-key catch-up. If it is already `<=` restored wm, first publish is allowed. Documented v1.

**(6)** GC grace is not a pin refcount. Dropping `E > prev_E ∧ attempt != owner` is safe for in-flight WRO. Remaining race is event-time floor vs a long request.

#287 overlay, #288 lease, #289 GC, #290 WRO filter need to match this, not the previous comment.

anovv:
**Correction:** ship the `prev_E` WRO rewrite; **revert** any-attempt WO overlay (`a1607604`).

WO stays:

```text
attempt == me
OR (attempt == cp.attempt AND E <= cp_E)
```

Until restore: `me` only. `me | E <= cp_E` (any attempt) launders uncheckpointed rows from a dead attempt (A leftover `E=50` beats B replay `E=12`). Streaming durability is checkpoint + source replay, not a serving line. Idle keys after a second failover are a v1 `StateVersion` cost; WRO still sees them.

WRO (unchanged from the prev_E fix):

```text
E <= serving_E
AND (E <= prev_E OR attempt == serving.attempt)
```

`prev_E` is 0 in CQL (no unset). Empty serving ⇒ no pin. Writer `E = max(cp_E, serving_E, prev_E)`. GC remains `E > prev_E AND attempt != owner`.

Issue #169, opened 2026-08-01
Window recovery can drop late rows after checkpoint restore


anovv:
## Summary

Window checkpoint recovery (and sometimes steady-state catch-up) can drop in-time rows as late. Reproduces on local single-worker and multi-worker / Kubernetes stress. After [#176](https://github.com/volga-project/volga/pull/176), this shows up as `assert_no_window_late_drops` failures (e.g. `test_local_single_worker_window_checkpoint_restore`).

## Evidence (original)

- Sequential stress: 2/20 local failures and 4/20 Kubernetes failures.
- Four parallel local shards reproduced immediately; three reported window-oracle mismatches and one hit the former 30-second stress timeout.
- Checkpoint diagnostics recorded a consistent source/window snapshot before recovery, then `dropped_late_rows=5` at window watermark `50000` after recovery.
- The final oracle failure was a missing row at timestamp `46000`, which should not arrive after that watermark in a single ordered attempt.

## Evidence (updated, post-#176)

- Late-drop accounting confirms drops on the restored attempt (not only sink oracle mismatch).
- Example single-worker failure: `window dropped 25 late rows: [("Window_1_1", 30, 10), ("Window_1_0", 25, 15)]`.
- Instrumented run on `debug/window-late-drop-cut`:
  - Restore set `watermark_frontier=Some(15000)` (checkpoint cut).
  - Window then advanced to `25000`, and **after that** dropped keyed batches with `batch_ts=[21000,25000]` at `frontier=Some(25000)`.
  - Drops are whole 5-row per-key batches (KeyBy split of datagen `batch_size=20` / 4 keys).
- Failure is flaky (race); single-worker Abrupt kill still fails often — **no surviving peer / zombie worker required**.

## Current hypothesis (revised)

**Not** primarily cross-attempt stale data. Attempt-id on all messages is the wrong fix for this failure mode: a replaced worker cannot reconnect as a zombie, and the single-worker case has no peer backlog.

Root cause: **watermark re-assignment on the Window after KeyBy splits a batch**.

1. Auto watermark assign is placed on the **Window** (closest window to each source in `logical_graph`), not on the source.
2. With assigner enabled, stream-task preprocess **drops** upstream watermarks and rebuilds WM from each **keyed** data message (`ooo=0` → WM = max event time in that message).
3. KeyBy turns one source batch into many keyed messages (first returned, rest pending).
[...]

PR #171, opened 2026-08-01
Instrument window recovery late-row race

anovv:
## Summary
- delay checkpoint propagation acknowledgement until barrier processing and forwarding complete
- add opt-in source/window checkpoint diagnostics and late-row-drop reporting
- document the reproduction and proposed attempt-fencing fix in #169

## Test plan
- `VOLGA_CHECKPOINT_DIAGNOSTICS=1 scripts/test stress --env local --test test_local_multi_worker_window_checkpoint_restore --runs-per-shard 20 --shards 1`

Depends on the checkpoint runner PR for the reusable stress command.


anovv:
not needed anymore

PR #179, opened 2026-08-03
Place watermark assign via event-time plan lineage

### LINKED ISSUE #169: Window recovery can drop late rows after checkpoint restore

anovv:
## Summary
- Resolve window ORDER BY event-time column lineage in the DF plan and attach `watermark_assign` (`TimeHint::ColumnName`) on the defining Source/Projection, before KeyBy.
- Remove closest-window BFS placement and Window-side assign/derive paths so Window only min-merges upstream watermarks.
- Fixes [#169](https://github.com/volga-project/volga/issues/169): keyed fragments no longer advance global WM and late-drop sibling keys from the same batch.
- Raise inmem window checkpoint inline limit to 8 MiB (WM fix retains more state) and log checkpoint size on each inmem snapshot.

## How assign placement works
1. Find windows in the DF plan that need event-time watermarks.
2. Read the window `ORDER BY` column — that is the event-time column (must be a real column).
3. Walk backward through the plan to where that column is **defined**:
   - Filter / rename alias (`timestamp AS event_time`) → keep walking with the inner column.
   - Computed expression (e.g. `timestamp + INTERVAL '0' MILLISECOND AS event_time`) → **stop at that Projection** (it creates the column the assigner will read).
   - Table scan → stop at that **Source**.
   - Join / multi-input → unsupported for auto placement for now.
4. Attach `watermark_assign` (`ColumnName` + ooo/idle) on that Source/Projection in the Volga graph.
5. Validate the assign site is **upstream of the window’s KeyBy** so WM is generated on the full (pre-split) batch.
6. Window has no assigner — it only min-merges upstream WMs.

## Test plan
- [x] Lineage unit tests (source / rename / computed / join error)
- [x] Planner window placement asserts (assign on Source, absent on Window/KeyBy)
- [x] Watermark streaming e2e late-drop == 0
- [x] `test_local_single_worker_window_checkpoint_restore` + multi-worker
- [x] Local stress 10× both window restore tests
- [ ] CI green after 8 MiB inmem limit bump

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -10,15 +10,32 @@ on:
   workflow_dispatch:
     inputs:
       profile:
-        description: "Test profile (default=unit+inprocess; all=default+docker+kube)"
+        description: "Test profile (default=unit+inprocess; all=default+docker+kube; kube-stress / inproc-stress)"
         type: choice
         options:
           - default
           - docker
           - kube
           - all
-          - stress
+          - kube-stress
+          - inproc-stress
         default: default
+      stress_test:
+        description: "Exact test name for inproc-stress (ignored otherwise)"
+        type: string
+        default: test_local_multi_worker_window_checkpoint_restore
+      stress_machines:
+        description: "CI runners (machines) for kube-stress / inproc-stress"
+        type: string
+        default: "3"
+      stress_shards_per_machine:
+        description: "Parallel stress processes per machine (kube default 1 = one Kind cluster)"
+        type: string
+        default: "1"
+      stress_runs_per_shard:
+        description: "Iterations per shard for kube-stress / inproc-stress"
+        type: string
+        default: "10"
 
 concurrency:
   # PRs cancel superseded runs; scheduled stress must not cancel itself (jobs can exceed 1h).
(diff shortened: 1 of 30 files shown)


anovv:
Closing in favour of https://github.com/volga-project/volga/pull/183

PR #182, opened 2026-08-03
Disable interval checkpoints during StopSources drain

anovv:
## Summary
- On `StopSources`, disable new interval checkpoints and abort any in-flight CP so drain can finish (`acks=0/N` hang)
- Re-enable checkpoints when a new execution attempt starts
- Make `scripts/stress-test` tolerate missing `rg`

Depends on #181.

## Test plan
- [ ] `scripts/test stress --env local --test test_local_multi_worker_window_checkpoint_restore --runs-per-shard 10 --shards 4`
- [ ] Confirm no hang with `acks=0/N` after `StopSources`

### DIFF
--- a/scripts/stress-test
+++ b/scripts/stress-test
@@ -159,6 +159,10 @@ done
 
 echo "Stress logs: $LOG_DIR"
 if [[ "$failed" -ne 0 ]]; then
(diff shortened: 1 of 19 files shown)


anovv:
Folded into #186 (stack is now #183 → #181 → #186). StopSources abort/disable CP lives on the drain-run-modes tip.

Issue #191, opened 2026-08-05
Redesign Worker as a dedicated actor


anovv:
## Summary

Redesign `Worker` as its own actor (same style as `StreamTaskActor` / transport backend actors), instead of a shared struct with interior mutability that fans out `ask`s to task actors.

## Why

Today the worker process hosts:

- a gRPC `WorkerService` that must stay responsive (heartbeat, `get_worker_state`, checkpoint/drain RPCs)
- a `Worker` that owns task/runtime/backend state and coordinates lifecycle by messaging nested actors

Those two concerns were previously glued together with `Arc<Mutex<Worker>>` (tokio mutex). Holding that lock across task-actor `ask`s starved concurrent RPCs — e.g. `close_worker_tasks` / `get_worker_state` blocking heartbeat and finish. We mitigated that with split “clone handles under lock, await outside” APIs (`WorkerStatePoll`, `task_signal_handles`), then replaced the outer mutex with `Arc<Worker>` + `std::sync::Mutex<WorkerInner>` so the compiler rejects holding the lock across `.await`.

That interior-mutability refactor is an interim fix. The structural smell remains:

1. **Worker is an actor in all but name.** Lifecycle (`configure` / `start` / `run` / `close` / `reset`), state polling, checkpoint barrier injection, and fatal quiesce are message-driven coordination over nested actors. Encoding that as mutex + method calls forces every call site to manually snapshot `ActorRef`s and re-enter after awaits.
2. **Coarse shared state + nested asks does not compose.** Any new RPC that needs both “consistent worker metadata” and “talk to tasks” reintroduces the same pattern (or races if validation and mutation are split across lock scopes).
3. **An actor mailbox serializes worker commands naturally.** Heartbeat / state reads can be cheap messages (or read snapshots) without inventing lock protocols. Long-running work (start, close, barrier fan-out) stays on the worker actor’s turn or is explicitly delegated, without blocking the gRPC accept loop via a shared mutex.
4. **Clearer ownership and reset.** Reset/reconfigure today tears down runtimes under sync critical sections and carefully avoids Drop-while-entered. A worker actor with an explicit `Reset`/`Configure` protocol makes incarnation boundaries (execution attempt id, health bus, task set) first-class instead of “replace fields under a lock”.

## Proposed direction

- Introduce something like `WorkerActor` with messages mirroring `WorkerService` RPCs (and internal ticks for state poll / fatal handling).
[...]

PR #186, opened 2026-08-03
StopSources drain via shared AttemptPhase (fix PipelineFinished hang)

anovv:
## Summary
- Shared `AtomicAttemptPhase` for run loop + StopSources drain
- Abort in-flight checkpoints on drain; reject new starts while draining
- Keep poll/checkpoint RPCs **outside** `select` so ticks cannot cancel `get_worker_state`
- Gate checkpoint select arm on `AttemptPhase::Running`
- Unify task-status checks via `WorkerSnapshot::all_tasks_in`

Stacked on:
1. Worker concurrency fix
2. CheckpointCoordinator actor

## Test plan
- [ ] PipelineFinished after StopSources (no hang)
- [ ] Checkpoint settle / drain harness paths
- [ ] Failure recovery still aborts in-flight CP

PR #193, opened 2026-08-05
Make Worker a kameo actor for control-plane concurrency

anovv:
## Summary
- Turn `Worker` into a kameo `Actor` (`ActorRef` in `WorkerServiceImpl`)
- Control messages carry `execution_attempt_id`; fence checks live in the actor
- `Worker` shell over `Option<WorkerInner>`; health recreated per incarnation
- **Close:** signal the watch; if still mid-run, abort the run loop. Background runtime dispose so reset stays under master budget
- Parallel task-signal fan-out; upstream end without max WM finishes instead of panicking

Stacked on **#200** (MasterLifecycle actor / Finish intent).

Related: **#191**.

## Test plan
- [x] `scripts/test default` (unit + inprocess)
- [ ] local / CI inproc stress on window checkpoint restore


anovv:
Stack: **#193** (this) → #194 (checkpoint actor) → #186 (drain/AttemptPhase). Related follow-up: #191.

Issue #192, opened 2026-08-05
Redesign master control plane as actors (retire MasterState mutex bag)


anovv:
## Summary

Evolve the master from `MasterState` as a bag of shared `tokio::Mutex`es (+ cross-cutting atomics) toward **owned actors** with message-passing boundaries. End state: no co-owned mutable protocol/registry/checkpoint state that both the attempt loop and gRPC update under locks.

## Why

Same structural smell as worker-side coordination ([#191](https://github.com/volga-project/volga/issues/191)): shared mutable state + async IO/asks forces lock escape hatches, atomics for cross-task poking, and easy regressions (holding a mutex across `.await`).

The attempt run loop (`ExecutionAttempt::run`) is already actor-shaped. The debt is the shared facade around it (workers registry, config, journal, snapshots, `AtomicAttemptPhase`, formerly `Mutex<Checkpoints>`).

`CheckpointCoordinator` is the first extract on this path (checkpoint protocol + persist owned on one mailbox). This issue tracks continuing that pattern for the rest of the master control plane.

## Direction (incremental — not one mega-MasterActor)

1. Keep / grow **attempt loop** as attempt authority (phase, CP start/abort, recover, finish).
2. Shrink **`MasterState`**: registry + journal/snapshot as owned components or small actors; hold `ActorRef`s instead of mutex bags.
3. gRPC stays a thin adapter (`tell`/`ask`), not a co-owner of shared mutable state.
4. Keep **slow worker RPCs off** the control-plane mailbox — fan-out `get_worker_state` / configure / reset stay delegated; results come back as messages.

## What “no shared mutable state” means

- **No** domain objects like `Mutex<WorkerRegistry>` / leftover checkpoint mutexes updated from both the run loop and gRPC.
- **Yes** to `Arc` handles, immutable snapshots, and explicitly published read-only views.
- Narrow primitives only where measured — not a general shared `MasterState` kitchen sink.

## Related

- Worker-as-actor (same class of redesign): #191
- CheckpointCoordinator actor: landed as the first master-side extract (see recent commits on the drain/checkpoint stack)

anovv:
First extract — CheckpointCoordinator actor: https://github.com/volga-project/volga/pull/194. This issue tracks continuing master control-plane actorization beyond checkpoints.

anovv:
Superseded by #.

[...]

PR #194, opened 2026-08-05
Introduce CheckpointCoordinator actor

anovv:
## Summary
- Own checkpoint protocol + store IO on `CheckpointCoordinator` (kameo actor)
- Remove `Mutex<Checkpoints>` / deferred `PendingPersist` from `MasterState`
- Persist completes inside the actor turn

First master-side extract toward actorized control plane — **#192** (same class of redesign as worker #191).

Depends on worker concurrency PR.

## Test plan
- [ ] `cargo test --lib runtime::master::checkpoint::`
- [ ] Checkpoint / recovery paths in harness


anovv:
Stack: #193 → **#194** (this) → #186. Related follow-up: #192 (and worker-side #191).

Issue #211, opened 2026-08-08
Observability: Flink-comparable metrics, state size, and worker RSS


anovv:
## Goal

Grow Volga metrics beyond throughput / path latency / backpressure-as-queue-fill so we can run Flink-comparable benches and debug WO + checkpoints. Prefer **engine-semantic** metrics in Volga; leave host/container CPU & cgroup memory to platform monitoring.

**Already done (not in scope):** scrape correctness (#208), worker-owned scrape (#206/#209), path latency rename (#207/#210).

## Baseline (keep)

- Task/op throughput counters + poll gauges
- `volga_stream_task_path_latency` (`recv − ingest`; omit when ingest absent)
- **Channel fill gauges** (keep): `tx_queue_size` / `tx_queue_rem` / `backpressure_ratio` per peer — answers *where* congestion is. Note: ratio is queue occupancy at write time, not “time backpressured”; gauges can go stale when a task is idle.
- Per-operator hist merge → percentile gauges (no fake pipeline-wide E2E latency)

### Current backpressure gap

`DataWriter` already returns per-send wall time (`backpressure_time_ms` = duration of `write_message`, including wait on `send().await` / timeouts), but the stream task **discards** it. That alone is not Flink-like. Prefer a task run-loop time budget (below) over a standalone send-wait hist.

### A. Progress & recovery

| Metric | Definition / notes |
|---|---|
| Watermark lag | Event-time progress vs wall (per task → op) |
| WM / barrier propagation delay | Graph-wide: stamp at inject/source, observe downstream. First-class — separate “slow op” from “stuck alignment / transport” |
| Checkpoint | Duration, align wait, payload **bytes**, success/fail — overlaps [#148](https://github.com/volga-project/volga/issues/148) |

### B. Task time budget (Flink-style) + worker RSS

| Metric | Definition / notes |
|---|---|
| `busyTimeMsPerSecond` (or equiv.) | Task doing work (operator / internal), not waiting on IO |
| `idleTimeMsPerSecond` | Waiting on **input** |
| `backPressuredTimeMsPerSecond` | Waiting on **output** (blocked in send / no buffer space) |
| Worker RSS | `volga_worker_rss_bytes` (+ optional virtual). Correlate with state size; **not** a substitute for cgroup/kube metrics |

[...]

PR #227, opened 2026-08-12
obs: busy/idle/backpressured ms-per-second (#211)

anovv:
## Summary
- Flink-style per-task `busy` / `idle` / `backpressured` ms-per-second gauges via `TaskTimeMetrics` + `report_task_time_metrics`
- **Idle** = input wait; **backpressured** = exclusive tx-queue wait via `BackpressureTracker` (same `DataWriter` / `BatchSender` path as BP ratio + queue gauges; `set_backpressure_tracker`); **busy** = residual wall time so the three sum ≈ 1000 ms/s
- Unit tests for ratio, blocked-time bounds, parallel exclusive accounting, and timeout cancel

Depends on #226 (checkpoint metrics).

## Test plan
- [x] `scripts/test unit --filter 'test(batch_channel)'` (5 passed)
- [ ] `scripts/test unit`
- [ ] Smoke a pipeline under load and confirm busy+idle+bp ≈ 1000 and BP ratio rises when queues fill

### DIFF
--- a/src/runtime/collector.rs
+++ b/src/runtime/collector.rs
@@ -65,22 +65,22 @@ impl Collector {
         partitions.iter().map(|partition_idx| self.output_channels[*partition_idx].clone()).collect()
     }
 
-    async fn write_message_to_channels(&mut self, message: &Message, channels_to_send: Vec<Channel>) -> HashMap<Channel, (bool, u32)> {
-        // parallel write
+    async fn write_message_to_channels(
+        &mut self,
+        message: &Message,
+        channels_to_send: Vec<Channel>,
+    ) -> HashMap<Channel, bool> {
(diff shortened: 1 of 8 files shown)


anovv:
CI `test_local_single_worker_same_addr_restart_fenced_recovers` failure was a race: crash cancelled an in-flight `get_worker_state` (`Cancelled` was hard-fail), so recovery recorded a bare `StatePollFailure` before the rebound unbound peer could return a fence error.

Fix: treat `Cancelled` as retryable (like `Unavailable`). Also restored kill-restore mid-batch cursor stop so denser checkpoint lifecycle events cannot skip `AttemptRunning` (that was the separate #226 flake).

PR #233, opened 2026-08-13
Serve HTTP /metrics for Prometheus scrape

anovv:
## Summary
Bottom of the v1 long-run soak work ([plan](docs/longrun-soak-plan.md) PR **A**).

- HTTP `GET /metrics` from `prometheus_handle().render()` on worker and master.
- Listen only when `VOLGA_METRICS_BIND_ADDR` is set (kube: `0.0.0.0:9090`). Unset locally so multi-worker TestCluster does not bind the same port.
- `init_metrics()` runs on server start so master checkpoint series exist even without a scrape listener.
- kubevolga: metrics port 9090 on master/worker pods and Services, `prometheus.io/scrape` annotations, env injected.
- Adds `docs/longrun-soak-plan.md` as the soak/Flink-bench tracker.

## Test plan
- [x] `scripts/test unit --filter 'test(metrics_http_serves_prometheus_text)'`
- [ ] `scripts/test unit`
- [ ] After image rebuild: curl worker/master `:9090/metrics` in kube


src/runtime/observability/metrics/http.rs
```diff
@@ -0,0 +1,101 @@
...
+            eprintln!("metrics HTTP server on {addr} failed: {err}");
+        }
+    });
+}
+
+async fn serve_metrics(addr: SocketAddr) -> std::io::Result<()> {
+    let app = Router::new().route("/metrics", get(metrics_handler));
+    let listener = tokio::net::TcpListener::bind(addr).await?;
+    println!("[metrics] HTTP /metrics listening on {addr}");
+    axum::serve(listener, app).await
```
anovv: A custom axum app is not needed here. `metrics` itself does not speak HTTP, but `metrics-exporter-prometheus` 0.17 (already a dep; default features include `http-listener`) already serves Prometheus text via `PrometheusBuilder::with_http_listener`.

`install()` is the wrong API because `init_metrics()` fans the Prom recorder out with the snapshot registry (and optional TCP) — it cannot steal the global recorder. The intended path for that setup is:

```rust
let (prom_recorder, exporter) = PrometheusBuilder::new()
    .set_buckets(...)
    .with_http_listener(addr)
    .build()?;
// add prom_recorder to the existing FanoutBuilder
tokio::spawn(exporter);
```

That is the same scrape payload as `handle.render()`, plus the exporter's upkeep loop. A second port (9090 vs gRPC) is correct; axum is not. Axum is already used for `request_source`, so this is not a new dep — it is still a second HTTP stack next to the one compiled in via `http-listener`.

src/runtime/observability/metrics/http.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+async fn metrics_handler() -> impl IntoResponse {
+    let body = prometheus_handle()
+        .map(|handle| handle.render())
```
anovv: If this custom handler stays, call `handle.run_upkeep()` before `render()`. In 0.17.2 those are separate: the crate's HTTP listener runs upkeep on scrape; `render()` does not. Idle-timeout / cardinality cleanup will not happen otherwise.

src/runtime/observability/metrics/http.rs
```diff
@@ -0,0 +1,101 @@
...
+    let addr: SocketAddr = match raw.parse() {
+        Ok(addr) => addr,
+        Err(err) => {
+            eprintln!("{METRICS_BIND_ENV}={raw} is not a socket address: {err}");
+            return;
+        }
+    };
+    tokio::spawn(async move {
+        if let Err(err) = serve_metrics(addr).await {
+            eprintln!("metrics HTTP server on {addr} failed: {err}");
```
anovv: Bind/parse failures are `eprintln` + continue, so kube `start()` succeeds and Prom scrapes nothing. Same for the parse error a few lines above. Bind before spawn (or fail `MasterServer`/`WorkerServer::start`) so a bad `VOLGA_METRICS_BIND_ADDR` or port-in-use is a process failure, not a silent missing scrape.

anovv (commented):
Three issues on the scrape listener — inline on `http.rs`. Env-gated bind and kube port/annotation wiring look right.

src/runtime/observability/metrics/http.rs
anovv: Done in 9d9cc26: crate `with_http_listener` + fanout in `registry.rs`; bind/parse fail `start`; `http.rs` removed.

src/runtime/observability/metrics/http.rs
anovv: Done in 9d9cc26: crate `with_http_listener` + fanout in `registry.rs`; bind/parse fail `start`; `http.rs` removed.

src/runtime/observability/metrics/http.rs
anovv: Done in 9d9cc26: crate `with_http_listener` + fanout in `registry.rs`; bind/parse fail `start`; `http.rs` removed.

PR #234, opened 2026-08-13
Add Count sink for long-run soaks

anovv:
## Summary
- Add `SinkSpec::Count` / `CountSinkFunction`: increment `volga_sink_records_written` and drop the batch (no InMemory store, no payload retain).
- Harness backends start the gRPC InMemory store only when the pipeline sink actually needs it; Count/Parquet/Request are no longer rewritten to InMemory.

Part of the long-run soak plan (PR B). Independent of #233.

## Test plan
- [x] `./scripts/test unit --filter 'test(/count_sink|install_in_memory_sink/)'`
- [ ] Existing kube/inprocess smoke still uses InMemory and still starts storage


src/tests/support/cluster_harness/resources/kube.rs
```diff
@@ -359,14 +373,8 @@ fn wait_for_pipeline(pipeline_name: &str) -> Result<()> {
...
-        let pipeline_deleted = kubectl(&[
-            "-n",
-            "default",
-            "get",
-            "volgapipeline",
-            pipeline_name,
-        ])
-        .is_err();
+        let pipeline_deleted =
+            kubectl(&["-n", "default", "get", "volgapipeline", pipeline_name]).is_err();
```
anovv: Please revert rustfmt-only wrapping in this file (and the same in `docker.rs` / `local/smoke.rs`: lifecycle client lines, `wait_for_pipeline_deleted`, timeout `anyhow!`, etc.). Leave the `needs_in_memory_store` / optional-storage changes; those are the actual harness fix.

anovv (commented):
Count sink + gating store spawn/install on `pipeline_needs_in_memory_store` looks right. Please drop the rustfmt-only hunks in the harness files so the real change is obvious.

src/api/spec/connectors/sink.rs
```diff
@@ -60,6 +62,28 @@ impl SinkSpec {
...
+        matches!(self, Self::InMemoryStorageGrpc { .. })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn count_sink_roundtrips_and_does_not_need_store() {
```
anovv: useless test, delete

src/tests/support/cluster_harness/resources/kube.rs
anovv: Reverted rustfmt-only wrapping in kube/docker/local harness files in 860f1a4. Storage gating is unchanged.

src/api/spec/connectors/sink.rs
anovv: Deleted in 860f1a4.

PR #235, opened 2026-08-13
Add CRD placement fields and Kind role topology

anovv:
## Summary
- Extend `VolgaPodSpec` with `nodeSelector`, `tolerations`, and `affinity`; DeepCopy them (maps/slices/pointers no longer alias) and copy onto the master Pod and worker StatefulSet.
- Kind cluster is 1 control-plane + infra + 1 tainted `volga-worker` node. Sample CR selects those roles; required hostname anti-affinity is omitted so CI can pack multiple worker replicas on N=1. Operator prefers infra via soft node affinity.

Part of the long-run soak plan (PR C). Independent of #233 and #234.

## Test plan
- [x] `cd kubevolga && go test ./api/v1alpha1/ ./internal/controller/`
- [ ] Recreate Kind (`scripts/kube-test-env destroy && scripts/kube-test-env setup`) and confirm nodes have `volga.io/role` labels/taints
- [ ] Apply sample pipeline and confirm master lands on infra, workers on the tainted worker node


scripts/kube-test-env
```diff
@@ -51,9 +87,10 @@ setup() {
   if cluster_exists; then
     echo "using existing Kind cluster: $CLUSTER_NAME"
   else
-    kind create cluster --name "$CLUSTER_NAME"
+    kind create cluster --name "$CLUSTER_NAME" --config "$ROOT/kubevolga/hack/kind-multi.yaml"
   fi
-  kubectl_test wait --for=condition=Ready nodes --all --timeout=120s
+  kubectl_test wait --for=condition=Ready nodes --all --timeout=180s
+  apply_volga_node_roles
```
anovv: Drop `apply_volga_node_roles` and `--config kind-multi.yaml` from default `setup`. Kube tests should keep `kind create --name "$CLUSTER_NAME"` (1 node, no taints).

This helper is soak/Flink-bench infra: label/taint a multi-node cluster. It is not a kube-e2e concern. Leave it out of test scripts until that env step exists (or put the function in a soak script, unused here).

CI already failed here: `invalid taint spec: volga.io/role=worker=worker:NoSchedule` (should be `volga.io/role=worker:NoSchedule` if/when this runs). On an existing 1-node cluster the skip path plus sample selectors would leave workers Pending.

anovv (commented):
CRD fields + DeepCopy + `applyPodPlacement` on master/worker look right — that is this PR.

Kind role topology should not be wired into kube e2e. `scripts/test kube` packs everything on one Kind node today (`kind create` with no config). Taints/selectors need an infra vs worker split, which a 1-node cluster cannot provide. `apply_volga_node_roles`, default `--config kind-multi.yaml`, and sample CR selectors belong in soak/bench infra setup (around E/F), not `kube-test-env setup`.

That is also the kube CI failure: setup calls `apply_volga_node_roles` and dies on `volga.io/role=worker=worker:NoSchedule` before any test runs. Fixing the taint spec is not enough if this still runs on every test setup.

kubevolga/config/samples/volga_v1alpha1_pipeline.yaml
```diff
@@ -17,6 +19,14 @@ spec:
         cpu: "2"
         memory: "2Gi"
   worker:
+    # Required hostname anti-affinity is omitted so CI (1 volga-worker node) can pack replicas.
+    nodeSelector:
+      volga.io/role: worker
```
anovv: Please revert `nodeSelector` / worker tolerations on this sample. Kube e2e builds every CR from this file and only overwrites `pipelineSpec` + `workers.replicas`, so these fields apply to all smoke/checkpoint/recovery tests. Those tests pack on one unlabeled node; selectors would make pods Pending.

Put role selectors on a soak/bench manifest later, not on the test template.

kubevolga/config/operator/operator.yaml
```diff
@@ -17,6 +17,16 @@ spec:
         app.kubernetes.io/component: operator
     spec:
       serviceAccountName: kubevolga-operator
+      affinity:
```
anovv: Same split: infra affinity is soak topology. Preferred is a no-op on an unlabeled 1-node cluster, but kube tests do not need it. Leave the operator deploy as it is today; pin to infra when soak setup exists.

scripts/kube-test-env
anovv: Done in b4291a4: `kube-test-env setup` is `kind create --name` again; `apply_volga_node_roles` removed. `hack/kind-multi.yaml` stays for soak later.

kubevolga/config/samples/volga_v1alpha1_pipeline.yaml
anovv: Reverted sample nodeSelector/tolerations in b4291a4.

kubevolga/config/operator/operator.yaml
anovv: Reverted operator infra affinity in b4291a4.

PR #237, opened 2026-08-13
Add volga-longrun soak (Steady + KillAfterCheckpoint)

anovv:
## Summary
- Adds `volga-longrun soak` on top of `VolgaCluster`: same sliding RANGE window job, `SinkSpec::Count`, Datagen `run_for_s` unset (soak loop owns duration), **prod 30s** checkpoints (not `kube_test` 2s).
- Scenarios: **Steady** (run for duration, watch `PipelineFailed`) and **KillAfterCheckpoint** (wait N completed CPs, abrupt kill, `wait_for_kill_restore`, continue). SIGINT wraps the whole scenario then shutdown.
- Count sink is **#234** (merged), not a second copy. It still appears in this PR’s file list because the GitHub base is **#236** (`feat/test-utils`), which does not contain Count. Do not merge D with B.
- Harness starts the in-memory store only when the sink needs it (`pipeline_needs_in_memory_store` on the D path).

`volga-longrun bench` and Prom dump/oracles are **#239**. Empty `FlinkCluster` stub only.

Stacked on **#236**. Includes **#234** via merge. Do not merge with **#233** / **#239**.

## Test plan
- [x] `./scripts/test unit --filter 'test(/longrun::/)'` (kill parse + reject)
- [ ] `cargo run --bin volga-longrun -- soak --env local --duration-secs 120`
- [ ] `cargo run --bin volga-longrun -- soak --env local --scenario kill --duration-secs 180`
- [ ] `soak --env kube` on Kind (Count must not spawn `volga-test-storage`)

PR #238, opened 2026-08-13
Add Prom + Grafana soak dashboard

anovv:
## Summary
- Adds `kubevolga/hack/bench`: Prometheus (`5s` scrape of `prometheus.io/{scrape,port,path}` pods) + Grafana with the soak board provisioned as home.
- Prefers `volga.io/role=infra` (soft affinity). Not kube-prometheus-stack; not part of `scripts/kube-test-env`.
- Dashboard rows: job checkpoints, stacked busy/idle/backpressured ms/s, watermark lag p50/p99, throughput, window operator (empty until those series exist).

Rebased onto current **#235** (1-node Kind, no role taints). Still needs **#233** for the `/metrics` listener and scrape annotations — not folded in, so this PR stays bench YAML only. Until A lands, Prom will see zero Volga targets.

```bash
make -C kubevolga bench
kubectl -n volga-bench port-forward svc/grafana 3000:3000
```

## Test plan
- [x] `kubectl kustomize kubevolga/hack/bench` renders Namespace, Prom, Grafana, dashboard ConfigMap
- [ ] `scripts/test kube` on this branch (should no longer die in Kind setup)
- [ ] After #233+#235 on Kind: `make -C kubevolga bench`, confirm Grafana loads *Volga soak* and Prom has `volga_*` series while a pipeline runs
- [ ] Confirm Prom/Grafana land on the infra node when `volga.io/role=infra` exists

### DIFF
--- a/kubevolga/Makefile
+++ b/kubevolga/Makefile
@@ -3,7 +3,7 @@ SHELL := /bin/bash
 KUBECTL ?= kubectl
 KUSTOMIZE ?= $(KUBECTL) kustomize
 
-.PHONY: install uninstall deploy undeploy sample unsample render
+.PHONY: install uninstall deploy undeploy sample unsample render bench unbench
 
 install:
 	$(KUSTOMIZE) config/crd | $(KUBECTL) apply -f -
@@ -25,3 +25,10 @@ unsample:
 
 render:
 	$(KUSTOMIZE) config/default
+
+# Prom + Grafana on infra (kubevolga/hack/bench). Opt-in; not part of kube tests.
+bench:
+	$(KUSTOMIZE) hack/bench | $(KUBECTL) apply -f -
+
+unbench:
+	$(KUSTOMIZE) hack/bench | $(KUBECTL) delete --ignore-not-found=true -f -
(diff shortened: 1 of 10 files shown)


anovv (commented):
Kube CI failure is inherited from the **old** #235 parent (`62bc6ff`), not from the bench YAML. Setup still runs `kind-multi` + `apply_volga_node_roles` and dies on `volga.io/role=worker=worker:NoSchedule` before any test. #235 already dropped that (`b4291a4`). This branch does not contain it.

**Stack:** F needs A + C, not the longrun binary.

1. Rebase onto current #235 (`feat/crd-placement`) so kube e2e stays 1-node / no taints. That should unblock `scripts/test kube`.
2. Also stack (or wait for) #233. Prom here scrapes `prometheus.io/{scrape,port,path}`; those annotations live on A. C-only has no `:9090` listener and no scrape labels, so Prom will see zero Volga targets.

Do not put `make bench` / this kustomize into `kube-test-env setup`. Soft infra affinity is fine.

`master ← #235 (C) + #233 (A) ← this PR (F)`

#234 / #236 / #237 are a different stack (Count + test_utils + longrun).

PR #239, opened 2026-08-13
Add soak Prom dump and fail-the-process oracles

anovv:
## Summary
- Teardown dump of the **#238 board** series (`pipeline_id=~".*"`, window-operator row included) plus `lifecycle.json`.
- `--dump DIR` writes lifecycle without Prom; `--prom-url` adds `prom.json` and Prom oracles.
- Oracles fail the process: WM advancing (skip when not sending), lag p99 sustain (bound = window RANGE + OOO), checkpoint health (silence / failed), unexpected `WorkerFailure` / `PipelineFailed`, restore id + lag catch-up fail cases.
- `PipelineSpec.state.checkpoint` (`interval_ms` / `timeout_ms` / `retention`) lives **on this PR**. Master and soak waits/oracles read spec first, Prod consts as fallback. Heartbeats / RPC / kube health stay on consts. `Some(0)` is invalid and is **not** a disable (consts `0s` is).
- Soak job sets 30s / 60s / 1 (same as Prod). `--config` YAML (**#241**) does not write these fields.
- Regenerated `kube_pipeline_spec.schema.json` so operator validation documents `state.checkpoint`.
- Rebased onto current **#237**.

Stacked on **#237**. Prom dump needs **#238** + **#233**. `--config` YAML is **#241** on top of this branch.

```bash
volga-longrun soak --env local --dump /tmp/soak
kubectl -n volga-bench port-forward svc/prometheus 9090:9090
volga-longrun soak --env kube --dump /tmp/soak --prom-url http://127.0.0.1:9090
```

## Test plan
- [x] `./scripts/test unit --filter 'test(/longrun::/)|test(/api::spec::state::/)'`
- [ ] Kind soak with Prom: dump files exist; oracles pass on a healthy Steady run
- [ ] KillAfterCheckpoint: HeartbeatUnavailable during restore does not fail; a panic outside the window does
- [ ] Local `--dump` without `--prom-url` writes `lifecycle.json` only


anovv (commented):
G on E is the right stack. Teardown dump + fail-the-process oracles match the plan. Local without `--prom-url` still runs lifecycle oracles; Prom checks need #233+#238.

Inherits #237 (vendored #234, `run_for_s` reset on restore, SIGINT only on the tail sleep — now also skips dump/oracles).

**Dump ≠ board.** `board_queries()` is close to #238 but omits the window-operator row (`volga_wo_*`) and `pipeline_id`. Plan was dump the board series.

**`--dump` requires `--prom-url`** — cannot archive `lifecycle.json` alone.

**`sources_finished` is `PipelineFinished`.** Soak is streaming / hold-on-finish, so that may never fire. The real WM skip is `records_sent_rate > 0.1`, not that flag.

### Tests — keep / drop

**Keep** (kill-window and Prom thresholds are easy to get wrong):
- `heartbeat_during_kill_ok` + `heartbeat_after_restore_fails`
- `panic_outside_kill_fails`
- `stale_lag_while_sending_fails`
- `lag_p99_sustain_fails`
- `checkpoint_silence_fails`
- `parse_rejects_kill_flag_on_steady`

**Drop or merge** (low signal):
- `parse_soak_defaults` / `parse_equals_and_bench` / `parse_dump_and_prom` / `parse_soak_kill_kube` — keep one kill+dump parse + the reject; the rest is argv trivia
- `samples_sum_same_timestamp` — PromQL already `sum()`s; helper is unused for oracles on single-series queries
- `restore_id_and_lag_pass` — happy path only; does not catch a low restore id or lag that never comes down

**Missing (higher signal than the drops):**
- checkpoint `failed` increase outside the kill window
- stale lag while **not** sending must pass (Datagen ended)
- `restore_checkpoint_id` fails when id < killed-after
- `evaluate` with `prom=None` still fails on unexpected `WorkerFailure`

src/longrun/dump.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+        by_ts
+            .into_iter()
+            .map(|(ts, value)| Sample { ts, value })
+            .collect()
+    }
+}
+
+/// Board series dumped at teardown (same queries as the Grafana soak dashboard).
+pub fn board_queries() -> &'static [(&'static str, &'static str)] {
```
anovv: These are not the same as the #238 board. The board has `pipeline_id=~"$pipeline"` and a window-operator row (`volga_wo_late_dropped_rows_total`, maintain pruned, state bytes, ingest hist). Dump omits all of that. Either share one query list or drop the "same as the dashboard" claim.

src/longrun/oracles.rs
```diff
@@ -0,0 +1,443 @@
...
+        }];
+        let kill = KillWindow {
+            at_unix: 100.0,
+            restore_at_unix: 110.0,
+            after_checkpoint: 1,
+            seq_lo: 1,
+            seq_hi: 2,
+        };
+        assert!(unexpected_fatal(&cfg(), &events, Some(&kill)).is_ok());
+    }
```
anovv, line 372: This pair (during kill ok / after restore fail) is the high-signal lifecycle test. Keep both. `restore_id_and_lag_pass` below is happy-path only — replace it with fail cases (restore id too low; lag never ≤ bound after restore).

src/longrun/run.rs
```diff
@@ -137,10 +147,84 @@ async fn run_kill_after_checkpoint(
...
+
+    let mut prom_dump = None;
+    if let Some(prom_url) = spec.prom_url.as_deref() {
+        println!("[volga-longrun] dumping Prom query_range from {prom_url}");
+        let dump = dump_board_series(prom_url, outcome.started_unix, end_unix, PROM_STEP)
+            .await
+            .context("prometheus query_range dump")?;
+        prom_dump = Some(dump);
+    } else if spec.dump.is_some() {
+        return Err(anyhow!("--dump requires --prom-url"));
```
anovv: `--dump` requiring `--prom-url` means you cannot persist `lifecycle.json` from a local soak. Allow dump of events without Prom.

anovv:
**Config / follow-up (do not fold the engine move into this PR).**

Keep this PR as G: dump + fail-the-process oracles. Do not insert a config PR between #237 and this one, and do not move checkpoint knobs onto `PipelineSpec` here.

**In this PR:** derive oracle/run waits from the job that actually ran. Today that is still consts (`RuntimeConstsProfile::Prod` / `master.checkpoint_interval` / timeout). Use those instead of `PROD_CHECKPOINT_INTERVAL = 30s` / `PROD_CHECKPOINT_TIMEOUT = 60s` in `oracles.rs` and the hardcoded `CHECKPOINT_WAIT` / `CHECKPOINT_IDLE` in `run.rs`. Same for lag: default from the soak window + `event_time` (RANGE 10s, OOO 0), not a magic 15s. No `--config` YAML and no `StateSpec` change here.

**Separate engine PR (sibling of the longrun stack, base master / current integration — not under #237):** `state.checkpoint` on `PipelineSpec` (`interval` / `timeout` / `retention`). Master reads spec first, consts as fallback. Heartbeats / RPC / kube health stay on consts. This is job API, not soak.

**After this PR (new PR on `feat/soak-oracles`):** soak/bench run YAML + `--config` (`env`, duration, scenario, launch, Prom/dump, oracle multipliers). CLI overrides the file. Once checkpoint-on-spec has landed, switch derivation from consts → `launch.pipeline.state.checkpoint` and allow `3x` / `1.5x_window`.

```
master ──► checkpoint-on-PipelineSpec          (engine, independent)
236 → 237 → 239 (this PR: dump + oracles, derive from consts)
                └──► soak --config YAML         (longrun only)
                       └──► oracles read spec.checkpoint when that PR is in
```

**Update the local soak plan** (`docs/longrun-soak-plan.md`): add these two PRs to the PR map, spell the three layers (consts / `PipelineSpec` / run YAML), and note that G derives from consts until checkpoint-on-spec lands. Do not merge the new engine PR with G, or the YAML PR with E.

anovv:
**Re-review: `state.checkpoint` on this PR (folded I).** Wiring is right: master interval/timeout/retention and soak waits/oracles use spec first, Prod consts as fallback. Heartbeats / RPC stay on consts. `Some(0)` rejected in `PipelineSpec::validate`. Dump follow-ups look addressed (`pipeline_id=~".*"`, `volga_wo_*`, `--dump` without Prom writes `lifecycle.json`, WM skip when not sending).

**Soak job still does not set the knobs.** `soak_window_launch_spec` leaves `CheckpointSpec::default()` (the new job test asserts that). Master and oracles therefore still read Prod consts. Putting the field on G does nothing for soak until someone fills it.

Set it on the **job** in this PR (30s / 60s / 1, matching today’s Prod), and assert those values — not `Default`. Do **not** wait for J, and do **not** have the run YAML write `launch.pipeline.state.checkpoint`. That mixes layers. J owns env/duration/scenario/oracles; checkpoint stays on `PipelineSpec`. Later J can point at a job file that already has `state.checkpoint`.

**Regen `kube_pipeline_spec.schema.json`.** `StateSpec` in the checked-in schema has no `checkpoint`. Extra properties are allowed so kube e2e will not fail, but operator validation will not document or reject `interval_ms: 0` (master `get_spec` will). This PR added `JsonSchema` on `CheckpointSpec` — run the schema bin.

**`Some(0)` vs consts `0s`.** Consts `0s` disables interval CPs (`optional_interval`). Spec `interval()` treats `0` as `None` (fallback to consts), and `validate` rejects `Some(0)`. You cannot disable from the job spec; a value that slips past validate becomes “use consts”, not “off”. Fine if that is intentional — say so. Do not document spec `0` as disable.

Plan line “J writes `launch.pipeline.state.checkpoint` from the run file” is the wrong split — fix that in the plan when you set the soak job fields here.

src/longrun/dump.rs
```diff
@@ -0,0 +1,262 @@
...
+        "prom_url": prom_url,
+        "start": start,
+        "end": end,
+        "step": step,
+        "series": series,
+    }))
+}
+
+#[cfg(test)]
+mod tests {
```
anovv, line 271: remove

src/longrun/dump.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+        by_ts
+            .into_iter()
+            .map(|(ts, value)| Sample { ts, value })
+            .collect()
+    }
+}
+
+/// #238 soak board exprs. Grafana uses `pipeline_id=~"$pipeline"`; dump uses `.*` (all).
+pub fn board_queries() -> &'static [(&'static str, &'static str)] {
```
anovv: name it prom queries

anovv:
docs/longrun-soak-plan.md - should not be tracked

anovv:
board_queries -> prom_queries.

Also those do not need to match whats in dashboard. In fact, can we make post-soak queries (or even better, just metrics names) configurable? So we can specify it without recompiling? we introduced soak yaml config in https://github.com/volga-project/volga/pull/241 - this will require some metrics to be mandatory (for oracles to have needed data) and we also can have a list of extra stuff if we want it. Not necessery to add yaml in this pr but make it configurable in code so we can factor it out in 241 easily

src/longrun/dump.rs
```diff
@@ -0,0 +1,262 @@
...
+    if body.status != "success" {
+        anyhow::bail!(
+            "prom query_range failed: {}",
+            body.error.unwrap_or_else(|| body.status.clone())
+        );
+    }
+    Ok(body)
+}
+
+pub async fn dump_board_series(
```
anovv: same renaming, do not use "board", just dump_prom_series

src/runtime/master/attempt/execute.rs
```diff
@@ -61,7 +61,9 @@ impl ExecutionAttempt {
             .run_health_poll(worker_ids, self.failure_tx.clone());
 
         let poll_interval = runtime_consts().duration(MASTER_STATE_POLL_INTERVAL);
-        let checkpoint_interval = runtime_consts().duration(MASTER_CHECKPOINT_INTERVAL);
+        let checkpoint_interval = self.pipeline.spec.state.checkpoint.interval_or(
```
anovv: why do we still keep cp related options in consts? This looks weird as they should have only one source -currently it is spec, remove it from runtime_consts, make sure if they are used anywhere in the codebase we wire values from spec instead. If its not on the spec we should panic

src/runtime/master/attempt/execute.rs
```diff
@@ -146,7 +148,9 @@ impl ExecutionAttempt {
             return;
         }
 
-        let checkpoint_timeout = runtime_consts().duration(MASTER_CHECKPOINT_TIMEOUT);
+        let checkpoint_timeout = self.pipeline.spec.state.checkpoint.timeout_or(
```
anovv: same for all cp related values that moved to spec

src/runtime/master/state.rs
```diff
@@ -255,7 +255,9 @@ impl MasterState {
                 task_index,
             })
             .collect();
-        let retention = runtime_consts().u64(MASTER_CHECKPOINT_RETENTION) as usize;
+        let retention = config.spec.state.checkpoint.retention_or(
```
anovv: same

anovv:
**Clarification on ripping `master.checkpoint_*` from consts** (updates the execute.rs / state.rs inline notes).

Spec is the only source. Drop `master.checkpoint_interval` / `timeout` / `retention` from the three consts JSON files and from `interval_or` / consts fallback in master + soak.

Do **not** panic at runtime if the field is missing. `PipelineSpec::validate()` requires the three fields (allow `interval_ms: 0` = off, same as today’s consts `0s` / `optional_interval`).

Env cadence is **not** a consts overlay anymore. Different callers set different values on the job:

- Soak (`soak_window_launch_spec`): **30s / 60s / 1 on local, docker, and kube.** Soak already wanted Prod CPs in every env.
- Local checkpoint suite (`checkpoint_recovery_launch_spec`): 500ms / 5s / 2
- Kube e2e checkpoint/smoke: 2s / 60s / 2
- Sample CR: explicit `state.checkpoint`

`RuntimeEnv` still picks the cluster backend only. Master must not branch on env for checkpoint interval.

Set the soak job fields in this PR (assert 30s/60s/1, not `Default`). Checkpoint-suite helpers can be updated in the same PR so tests keep their fast CPs after consts keys go away.

src/longrun/job.rs
```diff
@@ -30,3 +35,42 @@ fn clear_datagen_run_for_s(pipeline: &mut PipelineSpec) {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
anovv: remove

anovv:
Looks generally fine. Two test trims:

**Drop** `soak_job_is_count_sink_window_prod` in `job.rs` — it only snapshots the launch spec (Count, window SQL, 30s CP). No behavior.

**`oracles.rs` — keep fail cases and the kill-window pair; drop the rest.**

Keep:
- `heartbeat_during_kill_ok` + `heartbeat_after_restore_fails`
- `stale_lag_while_sending_fails` + `stale_lag_while_not_sending_passes` (skip when not ingesting)
- `lag_p99_sustain_fails`
- `checkpoint_silence_fails`
- `checkpoint_failed_outside_kill_fails`
- `restore_checkpoint_id_too_low_fails`
- `restore_lag_never_under_bound_fails`
- `evaluate_without_prom_fails_on_worker_failure` (local soak still fails closed)

Drop:
- `panic_outside_kill_fails` — same as `evaluate_without_prom` with extra wrapping

PR #241, opened 2026-08-15
Add volga-longrun soak --config YAML

anovv:
## Summary
- Soak `--config` is the spec: scenario, dump, Prom, oracles, job overlay (sql / checkpoint / parallelism / datagen / window).
- Count sink and Datagen (`run_for_s` unset) are fixed in code. Worker count is `parallelism / slots_per_node`.
- CLI is only `--config`, `--env`, and `--duration-secs`. No `--scenario` / `--kill-after-checkpoint` / `--dump` / `--prom-url`.
- Omit `--config` for the default window job (steady, local).
- Example: `kubevolga/hack/bench/soak.example.yaml`.

Stacked on **#239**. Do not merge with **#237**.

```bash
volga-longrun soak --config kubevolga/hack/bench/soak.example.yaml --duration-secs 120
```

## Test plan
- [ ] `./scripts/test unit --filter 'test(/longrun::/)'`
- [ ] `--config` launches with file scenario/job; CLI `--env` / `--duration-secs` override
- [ ] omitted `--config` uses the default window job


anovv:
Rebased onto current #239 (`88eb3a05`) — that part is correct. `3x_interval` is applied after `launch.checkpoint`, so it sees the file interval. Do not merge with #237.

**Drop `launch.checkpoint`.** Same note as on #239: checkpoint is a job field (`PipelineSpec.state.checkpoint`), not a run-file field. J owns env / duration / scenario / dump / Prom / oracle multipliers / `worker_count`. Writing `interval_ms` here duplicates (and will fight) the soak job once G sets 30s/60s/1.

`3x_interval` / `1.5x_window` stay — they **read** the resolved job interval / window. The example should omit `launch.checkpoint` and let the job spec supply 30s. If you need a different CP cadence, that is a different job file (`pipeline: …`), not a run overlay.

Still open from last review:
- `duration_secs: 3600` vs oracle `30s` / `3x_interval` — use `parse_duration` for the run (`duration: 1h` or `3600s`)
- file `kill_after_checkpoint: 0` is accepted; CLI rejects `0`
- file `scenario: steady` + `kill_after_checkpoint: 2` is ignored; CLI errors
- `1.5x_window` test hardcodes 15s — assert `1.5 * soak_lag_p99_bound()`

src/longrun/cli.rs
```diff
@@ -207,4 +246,33 @@ mod tests {
     fn parse_rejects_kill_flag_on_steady() {
         assert!(parse_args(["soak", "--kill-after-checkpoint", "1"]).is_err());
     }
+
+    #[test]
+    fn config_file_then_cli_override() {
```
anovv, line 147: not needed

src/longrun/config.rs
```diff
@@ -0,0 +1,278 @@
...
+        if n == 0 {
+            bail!("duration must be > 0: `{raw}`");
+        }
+        return Ok(Duration::from_secs(n));
+    }
+    bail!("invalid duration `{raw}` (use Nh, Ns, Nms, Nx_window, or Nx_interval)")
+}
+
+#[cfg(test)]
+mod tests {
```
anovv, line 474: useless

anovv:
i dont like this 1.5x_window and 3x_interval stuff and custom parsing. What we should do instead is to expose full pipeline spec as part of yaml - either inline or as a path to file. Right now we always default to soak_window_launch_spec() which uses lots of hardcoded stuff - cp options, query shape (window only), window-specific args (fixed window size), parallelism, reuses checkpoint_recovery_launch_spec which also hardcodes stuff - these are ok for fixed cp tests we already have but soak should be more configurable, we want to be able to customize params including query, parallelism, cp stuff, etc. So the way i see it soak spec should include configurable pipeline spec, which we should validate for soak-specific constraints when launched (e.g. verifying sink is count, etc, whatever is needed for soak to work as expected).

anovv:
Also in 239 we plan to make prom queries configurable - soak should expose them also as optional data we'd like to fetch for post processing (not sure how to best do it - raw queries or just metrics names and augment them to queries internally)

anovv:
**Clarification — current model (supersedes `launch.checkpoint` and `1.5x_window` / `3x_interval`).**

Drop the custom duration multipliers and do not write checkpoint from the run file. The run YAML carries a **pipeline** (inline `PipelineSpec` or a path). Soak validates at launch (Count sink, `state.checkpoint` set, whatever else the runner needs). Default if omitted can still be today’s window job.

Checkpoint cadence lives on that job spec (30s/60s/1 for soak). `3x` silence / lag bounds are ordinary durations in the oracles block (`90s`, `15s`), or they are derived in code from `launch.pipeline.state.checkpoint` + window — not a parser for `Nx_interval`.

Prom dump list: after #239 makes queries a value (mandatory oracle set + optional extras), the same list can appear in this file. Not bare metric names.

J still owns env / duration / scenario / dump / Prom / `worker_count` / oracles. CLI overrides those. Do not merge with #237.

`config_file_then_cli_override` / `parse_window_multiplier` go away with the multipliers. Keep one test that CLI `--env` / `--duration` override the file after you have a pipeline-in-YAML shape.

PR #250, opened 2026-08-21
Fix bench dump names, kill oracle, lag, and Grafana layout

anovv:
## Summary
- Prom dump/Grafana queried `*_total` counters; the metrics crate exports `volga_checkpoint_completed`, `volga_stream_task_records_sent`, etc. without that suffix (histograms still use `_bucket`).
- Kill scenario emits `TransportDisconnect` first; treat it as expected in the kill window (with HB / pod unhealthy).
- Bench datagen uses `ProcessingTimestamp` so watermark lag is `wall_now − event time`, not ~55 years from IncrementalTimestamp starting at 1s.
- Grafana: pipeline / worker / task variables; split job health vs duration; task time is three series for the selected task; window ingest vs state size are separate panels.

## Test plan
- [ ] `make -C kubevolga bench` to reload the dashboard ConfigMap, then refresh Grafana
- [ ] Rebuild/run `volga-bench` with `--duration-secs 120` and Prom forwarded; dump `checkpoint_completed` / `records_sent_rate` should be non-empty; lag p99 should be seconds not years
- [ ] Kill scenario should not fail on TransportDisconnect during the kill window
- [ ] In Grafana, pick one task: Task time shows busy / idle / bp

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -14,14 +14,14 @@ fail-fast = false
 slow-timeout = { period = "120s", terminate-after = 1 }
 fail-fast = false
 
-# Kind / kube (scripts/test kube).
+# Kind / kube (scripts/test kube) and scripts/stress-test. Must exceed the
+# harness kube kill-restore budget (recovery_started+replacement+
+# attempt1_running ≈ 300s) so nextest does not mask LifecycleOracle
+# diagnostics on slow replaces (4-node Kind, mid-flight kill after a safe CP).
 [profile.kube]
-slow-timeout = { period = "90s", terminate-after = 1 }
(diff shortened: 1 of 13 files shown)


anovv:
kube-tests timed out at nextest **90s** on `test_kube_mid_flight_checkpoint_kill_after_safe_restores_prior` after the kill (`safe=1`, `in_flight=2`). The other 12 kube tests passed; sibling kill-restores on the same 4-node Kind run were 15–45s.

That test waits for CP1 complete + CP2 start before the replace, so 90s was below the harness kube wait budget (~300s) and nextest ate the oracle error. Same floor as stress: kube profile is now **360s**.

PR #251, opened 2026-08-24
Unlimited bench rate and per-edge queue gauges

anovv:
## Summary
- Datagen `rate: null` is unlimited; omitted rate stays 200/s. `interval: 0s` disables interval checkpoints and skips checkpoint oracles.
- Event time is IncrementalTimestamp from submit wall-clock (replayable). Unlimited soak YAML at `kubevolga/hack/bench/unlimited.yaml`.
- Transport Grafana/dump: **tx queue remaining** and **rx queued records** (backend-agnostic). No shared-gRPC demux metric — that hop is going away.

Stacked on #250.

## Test plan
- [ ] `kubectl apply -k kubevolga/hack/bench` to reload the dashboard (avoid `make bench` unless you want Prom TSDB wiped)
- [ ] Grafana Transport row shows tx rem + rx queued; no demux panel
- [ ] `volga-bench --config kubevolga/hack/bench/unlimited.yaml` runs without checkpoint oracles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ go.work.sum
 # Local working notes (not for review)
 docs/metrics-redesign-plan.md
 docs/longrun-soak-plan.md
+docs/transport-improvements-plan.md
--- a/kubevolga/Makefile
+++ b/kubevolga/Makefile
@@ -27,8 +27,10 @@ render:
 	$(KUSTOMIZE) config/default
 
 # Prom + Grafana on infra (kubevolga/hack/bench). Opt-in; not part of kube tests.
+# Restart Prometheus so scrape-config ConfigMap changes are picked up.
 bench:
 	$(KUSTOMIZE) hack/bench | $(KUBECTL) apply -f -
+	$(KUBECTL) -n volga-bench rollout restart deploy/prometheus
 
 unbench:
 	$(KUSTOMIZE) hack/bench | $(KUBECTL) delete --ignore-not-found=true -f -
--- a/kubevolga/hack/bench/dashboards/volga-bench.json
+++ b/kubevolga/hack/bench/dashboards/volga-bench.json
@@ -7,7 +7,7 @@
   ],
   "timezone": "browser",
   "schemaVersion": 39,
-  "version": 2,
+  "version": 5,
   "refresh": "5s",
   "time": {
     "from": "now-15m",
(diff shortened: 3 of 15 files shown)


anovv:
## Rx queue metrics vs tx

Rx is **not** reported the same way as tx, and that part is right.

`DataWriter` keeps the `BatchSender`s and samples inline on `write_message`. The interesting tx state is “blocked on send,” so you are already on the write path (`rem ≈ 0`) and a ticker is unnecessary.

`DataReader::message_stream_with_control(self)` **consumes** the receivers into unfold streams, so there is nothing left to sample. The stall we care about is also off the recv path: dest queue filling while the operator is in `process()` / blocked on tx (“full + dest idle = not polling”). Publishing only on `recv` (or from `report_task_time_metrics`) misses that. For remote edges this is also a **different** queue than upstream `tx_rem` (egress vs ingress).

So: keep the cloned `Arc<AtomicU32>` and a 1s dest-side ticker. Event-driven rx, or putting the gauge on `BatchSender::send`, is not simpler — send only covers the local queue; remote ingress is filled by the backend.

The comment “stay live while blocked in `poll_next`” is a bit off. Waiting in `poll_next` is `recv`; queues should be ~empty. The ticker matters for process / tx-block.

## Simplify the machinery

`RxQueueSnapshot` / `RxQueuePublisher` are ceremony around that. Suggested cuts:

1. **Drop the public types.** `DataReader` already has the receivers; `StreamTask` already has `vertex_id` + `metrics_labels`. One method is enough, e.g. `queued_by_source() -> Vec<(String, Arc<AtomicU32>)>`, plus a small abortable 1s loop in `stream_task` (or a private helper). No `publish()` / `spawn_publisher()`.
2. **Don’t put `metrics_labels` on `DataReader`.** That only exists to feed the snapshot and forces `DataReader::new(..., None)` in tests. Labels belong at the publish site.
3. **`JoinHandle::abort()` on drop**, not `AtomicBool` + interval stop checks (waits up to 1s).
4. **`TransportBackend::with_metrics_labels(...)` will not compile** and is unused for rx. Client configs already get labels via `transport_cfg.set_metrics_labels`. Delete that call.
5. **`BatchReceiver::queued_records()` is unused**; only the `Arc` clone is needed.

Net: same gauges, same 1s sample while stalled, no DataReader constructor change, no extra types, no backend wiring.

anovv:
Landed the rx-queue cut: dest-side 1s ticker on `DataReader` (`queued_by_source` + abort-on-drop), no snapshot/publisher types, no labels on `DataReader`, and the unused `TransportBackend::with_metrics_labels` call is gone.

Issue #229, opened 2026-08-12
tcp based task-to-task transport backend


anovv:
current impl of grpc based backend uses single grpc stream channel between workers which is shared by tasks - head of the line block problem. We should switch to direct task-to-task and rely on raw tcp connection

anovv:
https://github.com/volga-project/volga/pull/254

anovv:
done

PR #252, opened 2026-08-24
Harden mesh harness: concurrent writers, order, local isolation

anovv:
## Summary
- Extend the existing many-to-many mesh harness (no second harness, no production change) so writers send concurrently on cloned `DataWriter`s — `TestDataWriterActor` is single-threaded and cannot isolate a blocked dest.
- Assert ordered delivery per `(writer, reader)` from the existing batch-index payload.
- Add a local slow-consumer case: one reader never `ReadMessage`, queue bound 2; other edges still finish and blocked sends do not return `false`. Remote slow-consumer is the same asserts, `#[ignore]` until #229 (shared gRPC HOL).

## Test plan
- [x] `./scripts/test inprocess --filter 'test(/backend_matrix/)'` — local/remote/mixed happy path + local slow-consumer pass; remote slow-consumer stays ignored
- [ ] CI inprocess suite


anovv:
## Review

Concurrent per-edge sends, ordered delivery, and local slow-consumer isolation are the right tests. A few cuts.

### Drop `TestDataWriterActor` and `TestDataReaderActor` together

`TestDataWriterActor` is why isolation could not work: a kameo actor serializes every `WriteMessage`, so one blocked dest stalls every dest. The harness already switched to cloned `DataWriter`s. The only remaining caller is `stream_task_actor_test.rs` (one sequential edge).

If the writer actor goes, drop the reader actor too. They are the same mailbox over `DataWriter` / `MessageStream`. Replacement:

- **Write:** `DataWriter::write_message` (already what the mesh does).
- **Read:** `DataReader::message_stream_with_control()`, then `stream.next().await`.
- **Pause:** keep the stream in a variable you never poll. Same effect as spawning an actor and never `ReadMessage`. Do not drop it — that closes the `BatchReceiver` and the paused send returns `false`. `_held_stream` is clearer than a live actor used as a lifetime holder.

Then delete `src/test_utils/transport.rs` (it is only these two actors) and inline the same calls in `stream_task_actor_test`.

### Paused `assert!(success)` is never observed

Paused send tasks are `abort()`ed and not joined. A panic from `success == false` is dropped. Until the write-path PR, `false` takes 5s×10 anyway, so abort after live traffic never hits it. After wait-until-queued, `false` means the channel closed — the case this assert is for.

After abort, await and surface panics; cancelled is the success path:

```rust
for handle in paused_sends {
    handle.abort();
    match handle.await {
        Err(e) if e.is_cancelled() => {}
        Err(e) if e.is_panic() => std::panic::resume_unwind(e.into_panic()),
        Ok(()) => panic!("paused dest send completed; queue bound should have blocked"),
        Err(e) => panic!("{e}"),
    }
}
```

### Smaller

- **`writer.start().await` is a no-op** (batcher is commented out). Drop it; same as the actor’s unused `Start` / `FlushAndClose`.
- **Remote slow-consumer topology is fine.** `RemoteOnly` offsets reader node ids, so `1` writer node + `1` reader node is still two backends / real remote edges. Mixed isolation omitted is correct.
- **Ignored remote case uses local timeouts** (`read_timeout: 5s`, `processing_wait: 50ms`). Harmless while `#[ignore]`; when #229 un-ignores it, use the happy-path remote waits (1s start / 10s read).
- **Order assert is the right contract** (per-writer index sequences, interleaving allowed). The old set-membership check + `TODO test ordered delivery` are done.
- **`apply_queue_size` is roundabout** (destructure → `new_*_with_queue` regenerates `channel_id`). Works because ids are a pure function of src/dst. Building the channel with the bound in `build_mesh_graph` would be shorter.

anovv:
Addressed the review:

- Deleted `TestDataWriterActor` / `TestDataReaderActor` and `src/test_utils/transport.rs`. Mesh and `stream_task_actor_test` use `DataWriter::write_message` / `DataReader::message_stream_with_control()`; paused dests keep `_held_stream` so the receiver stays open.
- After abort, paused send tasks are joined: cancelled is success, panics are resumed, completion panics.
- Dropped `writer.start()`. Queue bound is set in `build_mesh_graph` (`new_*_with_queue`). Remote slow-consumer uses the happy-path remote waits (still `#[ignore]` until #229).

anovv:
Rebased onto master after landing #249 / #250 / #251. Stack is still 252 → 253 → 254.

PR #254, opened 2026-08-24
Replace muxed gRPC remote hop with per-edge TCP

anovv:
## Summary
- Remote hop is one TCP connection per logical edge with a per-edge pump (`BatchReceiver` → socket). Local `BatchChannel` is unchanged. Flow control is the 8192 queue + TCP window; no credit, no user-facing backend type.
- Delete muxed gRPC (`message_stream` proto, `grpc_streaming` test, `TransportBackendType` / `resolve_transport_backend_type`). Production always TCP. Un-ignore remote slow-consumer — it passes.
- Add per-edge pump write-block ms/s and disconnect counters (`task_id`, `target_task_id`, no `backend=`). Dump PromQL + Grafana Transport panels; dashboard `version` 2.

Depends on #253 (write path) which depends on #252 (harness). Stacked: merge 252 → 253 → this. After image reload, unlimited bench should isolate (one full edge, others live) instead of silent-dying.

## Test plan
- [x] `./scripts/test inprocess --filter 'test(/backend_matrix/)'` — local/remote/mixed happy path + local and remote slow-consumer
- [x] `./scripts/test unit --filter 'test(/write_message_/) | test(/dump::tests/)'`
- [ ] CI unit + inprocess
- [ ] After image reload: unlimited kube bench does not silent-die; isolation fingerprint


anovv:
## Review

Architecture matches the plan: one TCP connection per logical edge, pump off the operator thread, local `BatchChannel` unchanged, muxed gRPC gone, remote slow-consumer un-ignored.

Bind-then-pumps, `TransportBackendType` / proto / `grpc_streaming` deleted, metrics labeled `task_id` + `target_task_id` (no `backend=`), dump PromQL matches the new panels.

## Issues

**1. Pump write-block does not sample during the block.**
`flush_write_block` runs after `write_frame` returns, or on a 100ms idle `recv` timeout. A multi-second TCP `write` (the fingerprint we want) leaves the gauge at the previous (low) value until the write completes. Wrap the write with a 1s ticker, or publish `write_start.elapsed()` while it is in flight.

Related: dest-not-polling also fills the TCP window, so you will usually see dest `rx` high **and** pump-block high. The plan’s “`rx` high, pump-block ~0 → not polling” will not show up with this hop. Still useful: `tx_rem≈0`, dest `rx≈0`, pump-block high → wire.

**2. `Message::from_bytes` can panic the ingress task.**
It `.unwrap()`s on a well-length-prefixed but corrupt payload. That is a panic, not `TransportDisconnect`. Map decode failure to `record_disconnect` + fatal, same as a short read.

**3. Ingress `tokio::spawn` tasks are not in `pump_handles`.**
`close()` sets `running=false` and aborts the listener + egress pumps. Ingress tasks blocked in `sender.send` (slow-consumer) never see `running` until send unblocks. Track them and abort on close, or `select!` send with a shutdown signal.

**4. Second connection for the same `channel_id` shares the queue.**
`senders.get(&channel_id).cloned()` lets two sockets interleave into one `BatchSender`. Take the sender out of the map (or reject) so a duplicate handshake is fatal/unknown.

## Smaller

- `TRANSPORT_GRPC_CONNECT_*` and `GRPC_MAX_MESSAGE_BYTES` in `tcp.rs` are leftover names. Keep the numbers; rename the connect keys (or a `transport.tcp_*` alias) so this is not still “grpc” in consts.
- Dashboard on this branch is only pump-block + disconnects (`version` 2). Tx rem / rx queued live in #251. After both merge, one Transport row with all four, single `version` bump.
- No `tcp.rs` codec tests is fine; the matrix is the test. A tiny frame max / handshake-unknown case would be cheap if you add anything.

Not in scope and fine: per-frame `flush`, 100ms recv poll, skipping an open-connections gauge, Collector `join_all`.

anovv:
Addressed the review:

1. **Write-block while in flight.** `write_frame` is wrapped in a 1s ticker that publishes `blocked_ns` before the write returns, so a multi-second TCP stall shows ~1000 ms/s instead of the previous window.
2. **Decode failure.** Corrupt payloads are `catch_unwind` around `Message::from_bytes` → `record_disconnect` + fatal (same as a short read). Did not rewrite the codec to `Result`.
3. **Ingress abort on close.** Accept tasks are held in `AbortOnDropHandles` so listener abort/shutdown cancels `sender.send` waiters.
4. **Exclusive handshake.** The `BatchSender` is `remove`d from the map; a second socket for the same `channel_id` is unknown/duplicate + fatal.

Smaller: connect retries are `transport.tcp_connect_*`; frame cap is a local 12MiB const (same number). Added `read_frame_rejects_oversize_len`. Dashboard still pump-block + disconnects until #251 lands.

Remote slow-consumer: a small burst can finish into the TCP window, so `Ok(())` after abort is allowed on remote; local still requires the send to stay blocked.

anovv:
Follow-up: the harden commit addresses the four review points correctly.

| Issue | What landed |
|---|---|
| Stale write-block gauge | `write_frame_sampled` keeps the `write_all` future pinned and ticks every 1s while it is in flight (immediate interval tick is consumed first, so you do not sample at t=0) |
| Decode panic | `catch_unwind` around `from_bytes` → disconnect + fatal |
| Orphan ingress tasks | `AbortOnDropHandles` in `listen_ingress`; `close()` aborting the listen task drops it and aborts children |
| Duplicate handshake | `Mutex<HashMap>` + `remove`; second socket is unknown/fatal |
| `transport.grpc_*` | `transport.tcp_connect_*` in consts + all three JSON profiles |

Remote slow-consumer un-ignored, muxed gRPC gone, one socket per edge. Looks good to merge after #252 / #253.

Leftovers, not blocking:

- `volga-worker.rs` still maps panics containing `[GRPC_BACKEND]` to `TransportDisconnect`. TCP does not panic that way anymore; the hook is dead. Fine to delete in a follow-up.
- Handshake / invalid UTF-8 / unknown channel still `report_fatal` without checking `running`. Shutdown is already tearing the listener down, so this is unlikely.
- Dashboard on this branch is still only pump-block + disconnects. Combine with #251’s tx rem / rx queued when those merge.

No need to re-litigate Collector `join_all`, per-frame flush, or the failure cascade (still worker-fatal by design).

anovv:
Picked up the leftovers:

- Deleted the `[GRPC_BACKEND]` panic → `TransportDisconnect` mapping in `volga-worker`. Panics are `Panic`; TCP disconnects already go through `report_fatal`.
- Invalid UTF-8 / unknown-or-duplicate handshake only `report_fatal` while `running`.

Dashboard still waits on #251 for tx rem / rx queued in the same Transport row.

anovv:
Dashboard is now based on #251 (v5 board) plus the TCP panels, **version 6**.

Transport row is 2×2: tx remaining, rx queued, pump write-block, disconnects (same `$pipeline` / `$worker` / `$task` filters). Dump extras include `tx_queue_rem` / `rx_queued_records` next to the pump-block and disconnect queries.

Apply with `kubectl apply -k kubevolga/hack/bench` (do not `make bench` just to refresh the board).

anovv:
docker-tests: `test_docker_smoke` hit `timeout waiting for docker pipeline` because the TCP hop parsed `worker-0:60052` as `SocketAddr` (IPs only). Compose DNS names used to work on the muxed gRPC hop.

Connect now uses `(host, port)` so tokio resolves the name. Unit test covers `localhost`; kube still uses pod IPs.

anovv:
Listen is no longer an open accept loop: after every expected ingress `channel_id` is claimed, the listener is dropped (pumps stay until close). Write-block/disconnect helpers are `report_write_block_gauge` / `increment_disconnects`.

Issue #159, opened 2026-07-30
Introduce KeyGroups as routing primitive


anovv:
## Why

Keyed routing is `hash % parallelism` (`pack_by_dest`). That bakes current `p` into every dest. [#263](https://github.com/volga-project/volga/issues/263) packed by dest and left `hash % p` as a temporary stand-in; the long-term contract is KeyGroups.

This ticket is the **primitive + live routing**. Rescale of keyed state is [#121](https://github.com/volga-project/volga/issues/121).

## Contract

One function, used for pack and later state assignment:

```
key_group = hash % max_parallelism
subtask   = key_group * p / max_parallelism
```

| Piece | Role |
|---|---|
| Hash | Job-lifetime, same `create_hashes` + seeds as `Key::new` ([#258](https://github.com/volga-project/volga/issues/258)) |
| `max_parallelism` | Constant for the job, `≥` any `p`. Default `= parallelism` so dests do not change vs today |
| Key group | `hash % max_parallelism` — does not change when `p` changes |
| Subtask | `key_group * p / max_parallelism` — contiguous ranges of groups per task |

Pack by **subtask** (at most `p` messages), never by key group. Do not also `HashPartition` the packed envelope with `hash % p`. Envelope extra stays `volga.target_subtask` (a subtask index).

`key_group_range(task_index, p, max_p)` is the inverse. Use the ceiling form so `kg ∈ range(subtask_of(kg))` when `p` does not divide `max_p` — **not** `i * max_p / p`.

Do **not** store key group on `Key`. `Key` stays hash + row; assignment always takes `max_p`.

### Primitive

New module (e.g. `src/common/key_group.rs`): `key_group_of`, `subtask_of`, `key_group_range`. Tests for the inverse, including uneven `p` vs `max_p`.

### Config

Job-wide `max_parallelism` (same as today’s single `parallelism`). Optional on `PipelineSpec` / kube spec; default `= parallelism`. Enforce `p ≤ max_p` at compile.

Thread: `PipelineSpec` → `PlanningContext` / `LogicalGraph` → `ExecutionGraph` or `ExecutionVertex` → `RuntimeContext`.

No per-operator `max_p` until per-node parallelism exists.

### Routing

- `pack.rs` `pack_by_dest`: dest = `subtask_of(key_group_of(hash))` instead of `hash % n`
[...]

PR #264, opened 2026-08-25
Pack KeyBy by dest and retire KeyedMessage

### LINKED ISSUE #258: KeyBy: reuse row hashes instead of Key::new recomputing them
## Summary

DataFusion `KeyByFunction::key_by` already `create_hashes` on the group columns, groups rows by hash, `compute::take`s each group, then:

```rust
// TODO this will re-compute hash - can we somehow change Key logic to use existing hashes?
let key = Key::new(key_batch).expect("should be able to create key");
```

`Key::new` runs `create_hashes` again (`src/common/key.rs`). With `batch_size: 20` and many unique keys this is ~20 one-row takes **and** a second hash per group on every source batch.

## Direction

- Let `Key` take a precomputed hash (or `Key::from_batch_and_hash`).
- HashPartition already uses `key.hash()` — keep that as the shuffle key; just do not hash twice.
- Optional later: avoid `take` for singleton groups (slice instead). Out of scope unless it stays cheap.

## Related

- [#220](https://github.com/volga-project/volga/issues/220) (multi-message drain) does not remove this per-batch KeyBy cost
- Code: `src/runtime/functions/key_by/key_by_function.rs`, `src/common/key.rs`

### LINKED ISSUE #263: KeyBy: pack by downstream task, not one KeyedMessage per key
## Problem

SQL / window planning builds `KeyByFunction::DataFusion`, not `ArrowKeyByFunction`. `DataFusionKeyFunction::key_by` (`src/runtime/functions/key_by/key_by_function.rs`) still groups by hash and returns **one `KeyedMessage` per distinct key**. `KeyByOperator` dribbles that `Vec` out one `poll_next` at a time via `pending_messages`. `HashPartition` then routes `message.key().hash() % p`.

Kind unlimited (`num_unique: 16384`, `batch_size: 20`) is typically 20 keys per source batch → 20 tiny KeyBy→Window messages (local queue or TCP). That multiplies frames, `Key::new` (recomputes hash), and Window `insert_batch` for the same ingest rate.

`ArrowKeyByFunction` does the same split. Planner never uses it (chained / `logical_graph` tests only). **Retire it** in this work: fold tests onto `DataFusionKeyFunction` (or `Custom`) and drop the enum arm.

## Fix: pack by downstream task

Not “one unkeyed batch to everyone” (that is broadcast). Still **hash every row**; stop **emitting one Message per key**.

In `KeyByOperator`, while the source batch is still one `RecordBatch`:

1. Evaluate PARTITION BY (as today).
2. `create_hashes` once; reuse those u64s ([#258](https://github.com/volga-project/volga/issues/258)).
3. Map each row to a dest: `key_group → subtask` (see rescale below). Temporary stand-in: `hash % p` (same as today, not rescale-safe).
4. `take` rows into **at most `p` batches** (skip empty dests).
5. Emit one Message per non-empty dest. Watermarks / barriers unchanged; collector still broadcasts them.

`p` is downstream channel count (`Collector.output_channels.len()`), not this vertex’s parallelism. Packing cannot live only in `HashPartition`: today KeyBy explodes first and the collector sees one key per send.

## Envelope: retire `KeyedMessage`, keep `Key`

**Drop `KeyedMessage` / `Message::Keyed`.** It exists so shuffle can `key.hash() % p` and Window/Reduce/Aggregate can skip splitting. After pack-by-dest the mailbox unit is multi-key (or one dest’s slice of keys). A single `Key` on that envelope is a lie.

Keep **`Key`** (hash, equality, checkpoint / store identity). Ingest splits the packed batch and calls `insert_batch(key, …)` / accumulators with a `Key`, not a `Message` variant. A local helper (`keys_in(batch) → (Key, RecordBatch)`) is fine. That is not `Message::Keyed`.

Do **not** keep the variant as unpack sugar: two ingest paths (`Keyed` vs Regular+split) forever, and someone will send a packed batch as `Keyed` with the first row’s key.

After packing, the hop is already routed: extras / `target_subtask`, or a `PartitionType` that reads the assigned dest. `Forward` is the wrong name (`num_partitions == 1` today). `HashPartition` on a packed envelope is wrong.

Other current `KeyedMessage` uses — convert, do not preserve:

- **Reduce output** re-wraps `Message::Keyed` so the next hop can hash without another KeyBy. Emit Regular with PARTITION BY columns still in the batch (or pack again).
- **WRO / request** is often one key per request. A one-row Regular batch is enough; `RequestRoute` uses extras, not `message.key()`.
- **In-memory sink** `keyed_buffer` keyed by envelope hash — derive from the key column.

## Rescale ([#159](https://github.com/volga-project/volga/issues/159))

Do not bake `hash % current_p` into the long-term contract. That moves every key when parallelism changes.

| Piece | Role |
|---|---|
| Hash | Job-lifetime, same `create_hashes` + seeds as `Key::new` |
| `max_parallelism` | Constant for the job, ≥ any `p` you will scale to. Stored on the graph and in checkpoints |
| Key group | `hash % max_parallelism` — **does not change** when `p` changes |
| Subtask | `key_group * p / max_parallelism` (contiguous ranges of groups per task) |

Pack rows by `subtask(key_group, p)`. Collector channel-id sort stays the tie-break for “subtask i → this edge” at a given topology; it does not replace key groups.

One function only: `hash → key_group → subtask`, used for pack, send, and later state assignment. Do not also `HashPartition` the packed envelope with `hash % p`.

Packing with `hash % p` is an acceptable **temporary** if this lands before KeyGroups; do not put that formula on the message type.

## vs [#220](https://github.com/volga-project/volga/issues/220)

Opposite sides of the hop. This ticket is KeyBy **output** (fewer messages). `#220` is Window **input** drain of several messages per poll.

`#220` alone still pays N frames, N `Key::new`, N flushes. This ticket still needs Window (and Reduce) to **split a multi-key batch** on ingest — same work `#220` / [#221](https://github.com/volga-project/volga/issues/221) want, on one Arrow batch instead of 20 mailbox items.

Do not close this as a duplicate of `#220`.

## Not this ticket

- Reusing hashes inside `Key::new`: [#258](https://github.com/volga-project/volga/issues/258)
- Forward vs RoundRobin on Source→KeyBy: [#142](https://github.com/volga-project/volga/issues/142)
- KeyGroups / rescale implementation: [#159](https://github.com/volga-project/volga/issues/159) (contract here must not fight it)
- TCP flush-every-frame: [#256](https://github.com/volga-project/volga/issues/256)

## Related

Umbrella: [#261](https://github.com/volga-project/volga/issues/261).

anovv:
## Summary
- KeyBy packs rows by downstream dest (`hash % p`) into Regular envelopes with `volga.target_subtask`; HashPartition reads that extra instead of hashing a per-key message ([#263](https://github.com/volga-project/volga/issues/263)).
- Window / WRO / Aggregate split packed batches locally. `Key::hash_arrays` / `Key::with_hash` reuse row hashes ([#258](https://github.com/volga-project/volga/issues/258)). Empty PARTITION BY / GROUP BY goes to dest 0 (one global key); hash collisions split by `Key` equality.
- Retire `Message::Keyed`, Arrow/Custom/Columns KeyBy, unused Reduce, and the unused `volga.key_fields` keyed-sink split.
- Window and WRO ingest per-key groups from a dest batch concurrently (`buffer_unordered(32)`) so remote store load/commit RTTs overlap.

## Test plan
- [ ] `./scripts/test unit --filter 'test(/functions::key_by::/)'`
- [ ] `./scripts/test unit --filter 'test(/operators::window::/)'`
- [ ] `./scripts/test unit --filter 'test(/operators::aggregate::/)'`
- [ ] `./scripts/test inprocess --filter 'test(/window/)'`
- [ ] `./scripts/test inprocess --filter 'test(/word_count/)'`
- [ ] Broader `./scripts/test unit` / `./scripts/test inprocess` as needed

Closes #263
Closes #258


src/runtime/functions/key_by/key_by_function.rs
```diff
@@ -1,382 +1,102 @@
...
-        // Step 1: Get key expressions based on the source type
-        let group_exprs = match &self.key_expr_source {
-            DFKeyExprSource::Aggregate(agg_exec) => agg_exec.group_expr().input_exprs(),
-            DFKeyExprSource::Window(window_exec) => window_exec.window_expr()[0].partition_by().to_vec(),
-        };
-        
+        let group_exprs = self.group_exprs();
         if group_exprs.is_empty() {
-            panic!("No group by expressions")
+            panic!("No group by expressions");
```
anovv: Commit 2 treats empty `PARTITION BY` as one global key in `split_by_key_exprs` (needed after DataFusion drops `PARTITION BY 1`). KeyBy still panics here.

Planner always inserts KeyBy in front of Window/Aggregate, so a SQL pipeline with `PARTITION BY 1`, `OVER (ORDER BY …)` (no partition), or `SELECT COUNT(*)` dies on the shuffle and never hits the split fix. The window smoke test only ingests into `WindowOperator` directly.

Pack the whole batch to dest `0` (same as a constant key). `split_message_by_key_fields` should treat missing/empty `volga.key_fields` as that global key — today it panics if the extra is empty.

src/runtime/operators/key_by/key_by_operator.rs
```diff
@@ -23,15 +25,17 @@ impl KeyByOperator {
...
             base: OperatorBase::new_with_function(key_by_function, config),
+            num_partitions: 1,
         }
     }
 }
 
 #[async_trait]
 impl OperatorTrait for KeyByOperator {
     async fn open(&mut self, context: &RuntimeContext) -> Result<()> {
+        self.num_partitions = context.parallelism().max(1) as usize;
```
anovv, line 41: [#263](https://github.com/volga-project/volga/issues/263) says pack with `Collector.output_channels.len()`, not this vertex’s parallelism. `HashPartition` then requires `dest < output_channels.len()`.

Planner currently sets the same `p` on every node, so this works today. If KeyBy and downstream ever diverge you either panic (`dest >= n`) or silently under-use dests. Wire dest to the Hash edge’s channel count, or document the same-`p` invariant here.

anovv (commented):
## Review

Pack-by-dest + retiring `KeyedMessage` matches [#263](https://github.com/volga-project/volga/issues/263): one Regular envelope per dest, `HashPartition` reads `volga.target_subtask`, ingest splits locally, `Key::hash_arrays` / `with_hash` reuse hashes ([#258](https://github.com/volga-project/volga/issues/258)). Watermarks stay on Broadcast in the collector. Reduce cleanup looks complete.

**Would not merge** until empty-key KeyBy is fixed (or those queries are rejected at plan time). Inline notes below.

Also: unit KeyBy tests only call `key_by(message, 1)`, so they never check dest packing / skipped empty buckets / HashPartition reading the extra. `word_count` with parallelism 2 is the real coverage — worth a small `pack_by_dest` test with `p > 1`.

src/runtime/functions/key_by/pack.rs
```diff
@@ -0,0 +1,210 @@
...
+    key_arrays: &[ArrayRef],
+    key_fields: Vec<Field>,
+) -> Vec<(Key, RecordBatch)> {
+    if batch.num_rows() == 0 {
+        return Vec::new();
+    }
+    let hashes = Key::hash_arrays(key_arrays, batch.num_rows()).expect("key hashes");
+    let mut groups: HashMap<u64, Vec<usize>> = HashMap::new();
+    for (i, &hash) in hashes.iter().enumerate() {
+        groups.entry(hash).or_default().push(i);
```
anovv: Grouping by hash only: two distinct keys with the same u64 hash become one group, and Window stores mixed partitions under the first row’s `Key`.

Packed ingest makes collisions merge **rows**, not just HashMap slots. `Key::eq` already falls back to record-batch compare — split should too after the hash bucket.

src/runtime/functions/key_by/key_by_function.rs
anovv: Valid — planner KeyBy ran before the ingest split, so `SELECT COUNT(*)` / `OVER (ORDER BY …)` never reached the global-key path.

Fixed in e089856: empty GROUP BY / PARTITION BY packs the whole batch to dest 0 with empty `volga.key_fields`. `split_message_by_key_fields` treats missing/empty extras as the global key.

src/runtime/operators/key_by/key_by_operator.rs
anovv: Valid observation; not changing dest math in this PR.

Planner still assigns the same `p` to KeyBy and Hash downstream, which is why `context.parallelism()` works today. Documented that invariant on `open` and left uneven rescale to #159 — KeyBy cannot see `collector.output_channels.len()` at pack time.

src/runtime/functions/key_by/pack.rs
anovv: Valid — hash buckets were merged into the first row’s `Key`.

Fixed in e089856: after grouping by hash, split further with the same record-batch compare `Key::eq` uses. Covered by `split_does_not_merge_distinct_keys_with_same_hash`.

PR #266, opened 2026-08-26
TCP transport: coalesce frame flushes (#256)

### LINKED ISSUE #256: TCP transport: do not flush after every frame
## Summary

`write_frame` in `src/transport/tcp.rs` does `write_all` of the length prefix + payload and then **`flush()` on every message**. Unlimited kube soaks (KeyBy split to 1-row messages, ~10k rec/s) turn that into ~10k flushes/s on each remote edge.

Flow control is already the 8192-record queue + TCP window. Per-frame flush is extra syscall / Nagle interaction, not correctness.

## Direction

- Flush when the edge queue is empty, the write buf is full, or after a small coalescing window — not after every `Message`.
- Keep length-prefixed frames; this is write-side batching only.
- Tests: existing TCP harness (`many_to_many` / `backend_matrix`) plus a soak signal that remote write-block / CPU drops without changing semantics.

## Related

- [#220](https://github.com/volga-project/volga/issues/220) (fatter messages) reduces frame count; flush coalescing still helps when messages stay small.
- Kind unlimited cell: RANGE 10ms, 16k keys, retractable aggs, source ~11k rec/s.

anovv:
## Summary
- Stop flushing the TCP write path after every length-prefixed frame (`write_frame` is write-only; `BufWriter` holds bytes until a real flush).
- Flush when the 8KiB write buf is full, or at most 2ms after the **first** unflushed byte (deadline is not reset by later writes). Handshake still flushes immediately.
- `TransportBackend::close` awaits egress pumps (then aborts on timeout) so the write buf is not dropped unflushed.
- Write-block metrics: `record_write_block_time` times each `flush`; `report_write_block_gauge` publishes the gauge.

## Test plan
- [x] `cargo test --lib -- tcp::` (one-frame deadline + trickle; plus pre-existing framing/listen)
- [x] existing TCP harness: `many_to_many` / `backend_matrix`
- [ ] Kind unlimited soak: remote write-block / CPU drops without changing semantics

Fixes #256

### DIFF
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -22,6 +22,5 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s",
-  "transport.tcp_flush_coalesce": "2ms"
+  "transport.tcp_connect_retry_delay": "1s"
 }
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -22,5 +22,6 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s"
+  "transport.tcp_connect_retry_delay": "1s",
+  "transport.tcp_flush_coalesce": "2ms"
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -22,6 +22,5 @@
   "worker.register_connect_timeout": "200ms",
   "worker.register_rpc_timeout": "500ms",
   "transport.tcp_connect_max_retries": 2,
-  "transport.tcp_connect_retry_delay": "50ms",
-  "transport.tcp_flush_coalesce": "1ms"
+  "transport.tcp_connect_retry_delay": "50ms"
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -22,5 +22,6 @@
   "worker.register_connect_timeout": "200ms",
   "worker.register_rpc_timeout": "500ms",
   "transport.tcp_connect_max_retries": 2,
-  "transport.tcp_connect_retry_delay": "50ms"
+  "transport.tcp_connect_retry_delay": "50ms",
+  "transport.tcp_flush_coalesce": "1ms"
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -22,6 +22,5 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s",
-  "transport.tcp_flush_coalesce": "2ms"
+  "transport.tcp_connect_retry_delay": "1s"
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -22,5 +22,6 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s"
+  "transport.tcp_connect_retry_delay": "1s",
+  "transport.tcp_flush_coalesce": "2ms"
 }
--- a/src/runtime/consts.rs
+++ b/src/runtime/consts.rs
@@ -39,6 +39,8 @@ pub const WORKER_REGISTER_CONNECT_TIMEOUT: &str = "worker.register_connect_timeo
 pub const WORKER_REGISTER_RPC_TIMEOUT: &str = "worker.register_rpc_timeout";
 pub const TRANSPORT_TCP_CONNECT_MAX_RETRIES: &str = "transport.tcp_connect_max_retries";
 pub const TRANSPORT_TCP_CONNECT_RETRY_DELAY: &str = "transport.tcp_connect_retry_delay";
+/// Max delay after the first unflushed TCP frame before the egress pump flushes.
+pub const TRANSPORT_TCP_FLUSH_COALESCE: &str = "transport.tcp_flush_coalesce";
 
 /// Explicit profile: `local_test` | `kube_test` | `prod`.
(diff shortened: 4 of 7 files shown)


anovv:
## Review + simpler loop

Direction is right: `write_frame` should not flush every message, handshake still flushes immediately so ingress can claim the channel, frames stay length-prefixed, `TCP_NODELAY` + `BufWriter` is the right pairing. The simplify commit (`ac93f17`) dropped the latency cap that #256 and the first commit both specified.

### Coalesce window is an idle timeout, not a deadline

`sleep(FLUSH_COALESCE)` is created fresh every `select` iteration, so any message within 2ms **resets** the timer. With `biased` preferring `rx.recv()`, a trickle faster than 2ms never hits the flush branch; bytes leave only when the 8KiB buf fills.

That is not “flush after a small coalescing window”. Small frames (watermarks especially) at a few hundred/s can sit until 8KiB fills — tens to hundreds of ms of extra delay, not 2ms. At ~10k rec/s the buf fills in a few ms, so the soak can look fine while moderate-rate paths do not.

The first commit had this right: `unflushed_since` + remaining `coalesce_wait` (deadline from the **first** unflushed byte). Do not reset on subsequent writes. Hardcode 2ms; no runtime const.

### `close()` abort can drop the write buf

Tokio `BufWriter` does not flush on drop. `TransportBackend::close` sets `running = false` and immediately `abort()`s pumps, so `if !running { writer.flush() }` often never runs. Previously every `write_frame` had already flushed. Await pumps after `running = false` (timeout, then abort).

### Tests

`write_frame_does_not_flush` is useful. `pump_egress_coalesces_many_small_frames` only checks that 512 watermarks arrive after `drop(tx)` (close-flush). It would pass with per-frame flush and never asserts the 2ms cap. Worth: one message does not arrive before the window; it does shortly after. Optionally a trickle faster than 2ms still flushes by the deadline.

PR body is stale (`transport.tcp_flush_coalesce`, “queue empty and window elapsed”).

### Simpler state machine

The 3-way `select` (`recv` / flush sleep / metrics tick, `if pending` / `if !pending`) plus the first commit’s `try_recv` drain / `flush_if_ready` are all one wait + one clock:

```rust
let mut since: Option<Instant> = None; // first unflushed byte

loop {
    if !running.load(Ordering::Relaxed) {
        let _ = writer.flush().await;
        return;
    }

    let wait = match since {
        Some(t) => FLUSH_COALESCE.saturating_sub(t.elapsed()),
        None => METRICS_TICK,
    };

    match tokio::time::timeout(wait, rx.recv()).await {
        Ok(None) => {
            let _ = writer.flush().await;
            return;
        }
        Ok(Some(message)) => {
            write_frame(&mut writer, &message.to_bytes()).await?;
            if writer.buffer().is_empty() {
                since = None; // BufWriter already flushed (buf full)
            } else {
                since.get_or_insert_with(Instant::now);
            }
        }
        Err(_) => {
            if since.take().is_some() {
                writer.flush().await?;
            } else {
                report_write_block_gauge(...);
            }
        }
    }
}
```

- Queue has data → `recv` returns immediately (natural drain, no `try_recv`).
- Burst fills 8KiB → `BufWriter` flushes itself, `since` clears.
- Idle or trickle → timeout fires when the **first** buffered byte is 2ms old.
- Empty and idle → 100ms tick for metrics and `running`.

Goes away: `biased`, the inverted pending branches, recreating `sleep(FLUSH_COALESCE)` every iteration, `write_queued_frames` / `flush_if_ready` / `try_recv`.

Pull the duplicated write-fail fatal into one helper (`report_write_fatal` from the first commit). Optional: only wrap `sample_write_block_metric` around `flush` and around `write_frame` when the write cannot fit in the remaining buf — otherwise every message builds a 1s `interval` for a memcpy.

anovv:
## Follow-up: comment is addressed; leftover cuts

The follow-up commit has the `timeout(recv)` + `unflushed_since` loop, close awaits pumps (then aborts) before tearing down ingress, and the tests actually pin the first-byte deadline. Trickle at 500µs spacing would fail on the old reset-sleep loop. Handshake still flushes immediately. I would not bring back `try_recv`, the 3-way `select`, or a runtime const.

### Leftover

`FLUSH_COALESCE` still says “after the last write if the edge queue is empty”. That is the old idle-timeout contract; the `pump_egress` docs are already correct.

### Biggest cut: drop `write_frame_maybe_blocked`

It exists only to skip the 1s metric timer on memcpy. Sample **`flush()` only** and call `write_frame` directly. Buf-full stalls then show up when `write_all` returns instead of on a 1s tick — rare next to the every-2ms flush. Do not sample every `write_frame` again (that puts the interval back on the 10k/s path).

### Smaller

- Give `report_write_fatal` a `what: &str` and use it for handshake (and connect if you want). The increment + `report_fatal` block is still copied twice above the loop.
- Handshake `async { write_frame?; flush }` can be two `if let Err` statements, or one tiny `write_frame_and_flush` used by handshake and the timeout-flush arm.
- Tests: `watermark_value` is unused in the soak; `running.store(false)` after `pump.await` is a no-op in three tests. The 512-frame soak does not assert coalescing (close-flush); the one-frame + trickle tests already do. Dropping the soak is optional.

anovv:
## Tests in `tcp.rs`

Not all of them are needed for this PR. Two pump tests lock the coalesce contract; the rest is either pre-existing framing/listen coverage or redundant with those two.

**Keep (coalesce contract)**
- `pump_flushes_one_frame_after_coalesce_window` — does not flush immediately, does flush at ~2ms.
- `pump_flushes_trickle_by_first_byte_deadline` — deadline vs idle-reset. This is the one that would have failed on `ac93f17`.

**Keep (pre-existing, not about flush)**
- `read_frame_rejects_oversize_len`
- `listen_stops_accepting_after_expected_edges_claimed`
- `connect_resolves_localhost_hostname` (DNS name vs `SocketAddr`; cheap)

**Drop**
- `write_frame_does_not_flush` — the one-frame pump test already fails if `write_frame` flushes.
- Do not bring back the 512-frame soak. Close-flush plus in-order delivery is not coalescing.

**Optional**
- `pump_flushes_when_running_cleared` — only coverage that `running = false` flushes instead of dropping `BufWriter`. Worth one test if you care about that close path; `TransportBackend::close` does not test it. Smallest set: drop it and rely on close awaiting the pump plus the one-frame deadline flush.

Minimum for #256: **one-frame + trickle**, plus the three older tests.

anovv:
## Rename `sample_write_block_metric`

It **records**, not a sample of flushes. Every `flush()` is timed in full: elapsed wall time goes into `blocked_ns`, and every 1s while still blocked we call `report_write_block_gauge` so a multi-second TCP stall shows up before `flush` returns. `report_write_block_gauge` is what actually `.set()`s the gauge (and no-ops until the local window is ≥1s).

“Sample” is misleading. Rename to `record_write_block_time`. Keep `report_write_block_gauge` for the `.set()`.

PR #267, opened 2026-08-26
Coalesce watermark emit on a processing-time interval

### LINKED ISSUE #260: Periodic watermark emit (coalesce on interval; drop per-batch assigner emit)
## Why

Assigner correctness is fine (`WM = max_seen − OOO`, source-only, min-merge, idle). Publish policy is not: we emit on every source batch that advances `max_seen`. That is punctuated/on-event. OOO does not slow emit rate (it only shifts the value). High-rate monotonic ingest (Kind unlimited soak) therefore sends ~one WM per source batch and Window `process_due` runs at ingest rate.

Flink’s production default is periodic (`autoWatermarkInterval` ~200ms processing time). We have no equivalent. Do **not** copy Flink `interval=0` (that **disables** periodic emit).

## Design (locked)

Keep: `out_of_orderness_ms`, `idle_timeout_ms`, `allowed_lateness_ms` (retention only), source-only assigner, emit **only if** `candidate > last_emitted`.

Add **`emit_interval_ms > 0`** on `WatermarkSpec` / `WatermarkAssignConfig`. One coalescer, no user-facing `on_advance`:

1. On each source batch: update `max_seen`. If `candidate` advanced **and** `now − last_emit ≥ interval` → emit now.
2. Else if advanced → mark dirty; mailbox/timer on the **source StreamTask** emits later.
3. Timer may fire when idle; **must not** send a WM unless candidate moved (no idle WM spam).
4. Always flush at EOF / max watermark.

Timer is required: data-path-only coalesce leaves a dirty WM stuck until the next record if the source goes quiet.

No named `OnAdvance` enum internally. Tests: fake clock, harness flush, or coalescer-only `Duration::ZERO` (not serde / not CRD). Do not document `0` as a user value.

### Defaults

| `ExecutionMode` | `emit_interval_ms` |
|---|---|
| `Streaming` | `200` |
| `Request` | `20` |
| `Batch` | EOF flush; interval unused |

Override via `event_time.watermark`.

## Implementation notes (not huge, not a drive-by)

Assigner math is small (`src/runtime/watermark/manager.rs`). The real work is **when** to `try_emit`:

- Sources inject in `StreamTask` after `poll_next` (`is_source` + `on_data_message` in `stream_task.rs`). The run loop is often blocked in source `poll_next`; a timer must `select`/mailbox with that poll (and stay correct next to checkpoint `try_recv`).
- `create_preprocessed_input_stream` also calls `on_data_message` and today only wakes on `input.next()`. If any assigner still lives there, that loop needs the same `select` + dirty flush. KeyBy/Window must not assign (`event_time_placement.rs`).
- Wire spec: `src/api/spec/event_time.rs`, `PipelineSpecBuilder`, `logical_graph` refresh like ooo/idle, kube schema `kubevolga/internal/controller/kube_pipeline_spec.schema.json`.

Finite tests that shut down the source often still close windows via **EOF max watermark**. Still add unit tests for: coalesce under high-rate advances, no WM when idle, dirty flush on timer without a following record, EOF flush. `src/tests/inprocess/watermark_streaming.rs` and window semantics should be checked for mid-stream emit assumptions.

## Out of scope

- Per-operator assigners / emit
- Auto interval from window length
- User-facing punctuated mode
- WO `process_due` / store lock ([#255](https://github.com/volga-project/volga/issues/255))
- Isolated WO bench ([#259](https://github.com/volga-project/volga/issues/259))

## Related

Kind soak: `kubevolga/hack/bench/unlimited.yaml` (`out_of_orderness_ms: 0` today). After this, streaming default 200ms should drop Window fire rate from ~source-batch rate to ~5/s without changing OOO.

anovv:
## Summary
- Periodic watermark emit ([#260](https://github.com/volga-project/volga/issues/260)): assigner coalesces on `emit_interval_ms` instead of emitting on every advancing source batch.
- One runtime default (200ms). Override via `event_time.watermark.emit_interval_ms` (Request benches that want 20ms set it on the spec). Dirty flush on source/preprocess timer and EOF/max. `Duration::ZERO` is tests-only (not a documented user value).
- Stacked on pack-by-dest ([#264](https://github.com/volga-project/volga/pull/264)).

## Test plan
- [ ] `./scripts/test unit --filter 'test(/watermark/)'`
- [ ] `./scripts/test unit --filter 'test(/event_time_placement/)'`
- [ ] `./scripts/test unit --filter 'test(test_window_query_to_request_mode)'`
- [ ] `./scripts/test inprocess --filter 'test(/watermark/)'`
- [ ] Broader `./scripts/test unit` / `./scripts/test inprocess` as needed

Closes #260


anovv (commented):
Same emit-interval default everywhere is fine.

Two leftover notes on barrier ordering vs a held watermark, and on covering the StreamTask timer path (inline).

src/runtime/stream_task.rs
```diff
@@ -458,80 +498,120 @@ impl StreamTask {
...
+                                    // Drop punctuated upstream WMs when this task assigns.
+                                } else if let Some(new_watermark) = watermark_manager.merge_watermark(watermark.clone()).await {
+                                    Self::set_watermark_lag_gauge(&vertex_id, new_watermark, labels.as_ref());
+                                    let mut wm = watermark.clone();
+                                    wm.watermark_value = new_watermark;
+                                    wm.metadata.upstream_vertex_id = Some(vertex_id.as_ref().to_string());
+                                    outgoing.push(Message::Watermark(wm));
+                                }
+                            }
+                            Message::CheckpointBarrier(barrier) => {
```
anovv: Checkpoint barriers can overtake a coalesced watermark.

This arm yields the aligned barrier with no `flush_pending`. Same on the source run loop: `checkpoint_receiver.try_recv` injects a barrier without flushing a dirty candidate.

Held `max_seen` then snapshots downstream window state behind event time the assigner already has. EOF/max already flush; barriers should too (source inject + preprocess align).

src/runtime/stream_task.rs
```diff
@@ -458,80 +498,120 @@ impl StreamTask {
...
-                                    &vertex_id,
-                                    new_watermark,
-                                    labels.as_ref(),
-                                );
-                                yield Message::Watermark(WatermarkMessage::new(
-                                    vertex_id.as_ref().to_string(),
-                                    new_watermark,
-                                    Some(Self::now_ms()),
-                                ));
+                    _ = Self::sleep_until_deadline(deadline) => {
```
anovv: Timer wiring here (and the matching source `select` on `poll_next`) is untested.

Assigner unit tests call `try_emit_due` with a fake clock; existing StreamTask preprocess tests use `WatermarkAssignConfig::new()` (`Duration::ZERO`), so they never arm `sleep_until_deadline`.

The failure #260 called out is a quiet source leaving a dirty WM stuck until the next record — that only happens if this `select` + deadline path is wrong. Worth one StreamTask/inprocess case with a non-zero interval: two advances, then idle, then a watermark with no following record.

src/runtime/watermark/manager.rs
```diff
@@ -15,6 +15,8 @@ use super::confg::{TimeHint, WatermarkAssignConfig};
 struct UpstreamWatermarkProgress {
     max_event_time_seen_ms: u64,
     last_emitted_wm_ms: u64,
+    dirty: bool,
```
anovv: `dirty` is redundant with unpublished progress (`candidate > last_emitted_wm_ms`). It is only set when we hold, and `emit` is the only clear.

Drop the flag. Collapse `on_data` / `try_emit_due` / `flush` onto one helper:

```rust
fn maybe_emit(progress, id, ooo, interval, now, force: bool) -> Option<WatermarkMessage> {
    let candidate = Self::candidate(progress, ooo);
    if candidate > progress.last_emitted_wm_ms && (force || Self::is_due(progress, interval, now)) {
        Some(Self::emit(progress, id, candidate, now))
    } else {
        None
    }
}
```

- data path: update `max_seen`, then `maybe_emit(..., force=false)` for that upstream
- `try_emit_due` = all upstreams, `force=false`
- `flush` = all upstreams, `force=true` (EOF / barrier / max)
- `next_emit_deadline`: filter `candidate > last_emitted` instead of `dirty`

Keeps assigner unit tests; rename `timer_does_not_emit_when_not_dirty` / `flush_emits_dirty_before_interval` to match.

src/runtime/stream_task.rs
```diff
@@ -436,6 +436,59 @@ impl StreamTask {
...
+        injected.set_upstream_vertex_id(vertex_id.as_ref().to_string());
+        injected.set_ingest_timestamp(Self::now_ms());
+        if let Message::Watermark(ref w) = injected {
+            Self::set_watermark_lag_gauge(vertex_id, w.watermark_value, labels);
+        }
+        Self::record_metrics(vertex_id.clone(), &injected, false, labels);
+        Self::send_to_collectors_if_needed(collectors_per_target_operator, injected).await;
+    }
+
+    async fn collect_flushed_watermarks(
```
anovv: `collect_flushed_watermarks` only covers flush. Timer still inlines the same merge-and-push loop (`try_emit_due` ~610), and the data path does it a third time for a single injected WM. Source repeats `send_source_assigned_watermark` for checkpoint flush (~890), timer (~933), and EOF (~1117).

Generalize:
- preprocess: `push_assigned_watermarks(manager, ..., wms)` used for `flush_pending()`, `try_emit_due()`, and the optional data-path inject
- source: `send_source_assigned_watermarks(..., wms)` wrapping the existing single-send helper

Optional: `WatermarkManager::{flush_pending,try_emit_due}` can drop the `now` argument and use `Instant::now()` internally (assigner keeps `now` for tests).

Leave the two `select`s (preprocess `input.next` vs source `poll_next`) as they are.

anovv (commented):
Original notes are addressed:
- Held WMs flush before the aligned preprocess barrier and before source `try_recv` inject.
- `stream_task_preprocess_timer_emits_held_watermark_without_next_record` covers the `select` + deadline path (idle, no following record). `stream_task_preprocess_flushes_held_watermark_before_aligned_barrier` covers barrier ordering.

Follow-up (inline): the coalesce policy can drop the extra `dirty` flag and the StreamTask emit/flush loops can share one drain helper. Do not unify the source vs preprocess `select`s — poll vs stream and send vs merge are different.

anovv:
## Executor: simplify coalesce emit (no behavior change)

Original review notes are already addressed (flush before barrier + StreamTask timer test). This follow-up is structure only.

### 1. Drop `dirty` in `WatermarkAssignerState`

`dirty` == unpublished progress (`candidate > last_emitted_wm_ms`). Collapse `on_data` / `try_emit_due` / `flush` onto one helper:

```rust
fn maybe_emit(progress, id, ooo, interval, now, force: bool) -> Option<WatermarkMessage> {
    let candidate = Self::candidate(progress, ooo);
    if candidate > progress.last_emitted_wm_ms && (force || Self::is_due(progress, interval, now)) {
        Some(Self::emit(progress, id, candidate, now))
    } else {
        None
    }
}
```

- data path: update `max_seen`, then `maybe_emit(..., force=false)` for that upstream
- `try_emit_due` = all upstreams, `force=false`
- `flush` = all upstreams, `force=true` (EOF / barrier / max)
- `next_emit_deadline`: filter `candidate > last_emitted` instead of `dirty`

Keep assigner unit tests; rename `timer_does_not_emit_when_not_dirty` / `flush_emits_dirty_before_interval`.

### 2. One drain helper in StreamTask

`collect_flushed_watermarks` only covers flush. Timer still inlines merge-and-push (`try_emit_due`), data path does it again for one injected WM. Source repeats `send_source_assigned_watermark` for checkpoint flush, timer, and EOF.

- preprocess: `push_assigned_watermarks(manager, ..., wms)` for `flush_pending()`, `try_emit_due()`, and optional data-path inject
- source: `send_source_assigned_watermarks(..., wms)` wrapping the existing single-send helper

Optional: `WatermarkManager::{flush_pending,try_emit_due}` can drop the `now` arg and use `Instant::now()` internally (assigner keeps `now` for tests).

Leave the two `select`s (preprocess `input.next` vs source `poll_next`) as they are.

PR #270, opened 2026-08-27
Pack in-mem window rows and drop the global store lock

### LINKED ISSUE #255: Window process_due parallelism: cap join_all and drop the global in-mem store lock
## Summary

Unlimited kube soaks (RANGE 10ms, retractable SUM/COUNT/AVG, 16k keys, p=4) still show **wm process p99 ~50–100ms** and **~1.0 CPU per worker** with a 2-core pod limit unused. Ingest histograms stay sub-ms. The hot path is `process_due`, not min/max rebuilds.

Two coupled problems in the Window watermark path:

1. `WindowOperator::process_due` `try_join_all`s a full due page with **no concurrency cap** (`TODO add concurrency limit` in `operator.rs`). In-mem `stream_due` page size is **256** triggers.
2. `InMemWindowStore` is a **single `tokio::sync::RwLock` over all keys + the global trigger `BTreeSet`**. `advance_key` loads under that lock, then `store_key_state` takes a **write** lock per key. 256 concurrent advances contend on one lock; extra Tokio threads on the task runtime cannot turn that into >1 core of useful work.

This is independent of [#220](https://github.com/volga-project/volga/issues/220) (multi-message ingest drain) and [#153](https://github.com/volga-project/volga/issues/153) (runtime-per-StreamTask). Serial `poll_next` per task stays; this issue is *how much parallel work one poll is allowed to start, and whether the store can actually run it*.

The in-mem backend comment already records the intent: one read lock is the WRO snapshot boundary. That is a snapshot protocol choice, not a reason to hold the same lock across 256 key evals on the soak hot path. In-mem is what Kind unlimited soaks actually run.

## Suggested direction

Do not make `poll_next` concurrent on one Window. Keep the mailbox.

**1. Cap `process_due` (small, matches existing TODO)**  
Semaphore around `advance_key` (start with `num_threads_per_task` or a small constant like 8). Stops 256-way future storms and write-lock stampedes. Will **not** by itself fill the second CPU if the store lock remains global.

**2. Release the lock before CPU eval (required for the cap to matter)**  
`load_raw` / `load_tiles` already clone under a read lock. Ensure `produce_slide` / `produce_rebuild` run with **no** store lock held. `store_key_state` / trigger updates stay short write sections.

**3. Per-partition (or sharded) locks for in-mem**  
`DashMap<PartitionKey, RwLock<PartitionState>>` (or shard by key hash) plus a trigger index that is not “scan the world under one lock.” WRO snapshot needs an explicit protocol (epoch, freeze, or consistent clone) so we do not silently break “one read lock = snapshot.”

**4. Do not overfit a durable backend**  
Rocks/object store will have their own concurrency. The in-mem reference backend should still be parallel enough that a Kind soak is not lock-bound, or soaks will keep measuring the lock.

## Acceptance

- `process_due` never `join_all`s an unbounded page
- Key eval CPU does not hold the process-wide in-mem lock
- In-mem snapshot/checkpoint/WRO behavior stays well-defined (test)
- Soak signal: wm process p99 drops and/or worker CPU can exceed ~1.0 on the 2-core limit without changing RANGE/keys (compare to current 16k / 10ms / retractable cell)

## Related

- Code: `src/runtime/operators/window/operator.rs` (`process_due`), `store/backend/inmem.rs` (`Arc<RwLock<InMemState>>`, `PAGE_SIZE = 256`)
- [#220](https://github.com/volga-project/volga/issues/220) ingest coalescing (fatter polls; complementary)
- [#153](https://github.com/volga-project/volga/issues/153) runtime-per-task (thread topology; complementary)

### LINKED ISSUE #257: In-mem window store: stop storing one Arrow RecordBatch per row
## Summary

`InMemWindowStore::commit_events` inserts **one `RecordBatch` per row**:

```rust
for (row, cursor) in cursors.into_iter().enumerate() {
    state.raw.insert(cursor, events.slice(row, 1));
}
```

`BTreeMap<Cursor, RecordBatch>` then forces `RowNav` to flatten many 1-row batches on every `process_due`. Arrow per-batch overhead (schema, validity, cache) dominates even when RANGE is 10ms (~10 rows/key).

This is separate from the **global `RwLock`** ([#255](https://github.com/volga-project/volga/issues/255)): sharding locks does not fix the layout.

## Direction

- Store **one (or few) batches per key** and index rows by cursor, or a packed column store keyed by `(partition, cursor)`.
- `slice(row, 1)` only if a caller truly needs a single-row view.
- Checkpoint encode/decode should round-trip the packed form.
- Tests: existing in-mem ingest/eval/checkpoint tests; soak: lower `wo_state_*_bytes` / ingest and process CPU at the same RANGE.

## Related

- [#255](https://github.com/volga-project/volga/issues/255) lock / `process_due` concurrency
- In-mem is what Kind unlimited soaks actually run

anovv:
## Summary
- **#257:** `InMemWindowStore` stores one cursor-ordered Arrow batch per key instead of one `RecordBatch` per row. Checkpoint encode/decode round-trips that packed batch.
- **#255:** `process_due` caps in-flight `advance_key` at 8 (`buffered`, ordered). The global store `RwLock` is gone: each partition has its own lock, eval CPU runs after loads clone out, and WRO `load_window_data` is one partition read (raw + tiles together).

## Test plan
- [ ] `cargo test --lib runtime::operators::window`
- [ ] Existing in-mem ingest/eval/checkpoint tests (packed-batch asserts + concurrent partition writes)
- [ ] Compare `wo_state_*_bytes` and wm process p99 / worker CPU vs current 16k keys / 10ms RANGE / retractable cell

Fixes https://github.com/volga-project/volga/issues/257
Fixes https://github.com/volga-project/volga/issues/255


src/runtime/operators/window/store/backend/inmem.rs
```diff
@@ -88,9 +185,26 @@ impl InMemWindowStore {
...
+    }
+
+    fn write_part<R>(
+        &self,
+        partition: &PartitionKey,
+        f: impl FnOnce(&mut PartitionState) -> R,
+    ) -> R {
+        let slot = self.partitions.entry(partition.clone()).or_default();
+        let mut state = slot.write();
+        f(&mut state)
```
anovv, line 210: **Requested change:** `entry().or_default()` keeps the DashMap shard exclusive for the whole `f()` — including `PackedRaw::insert` concat + `take_record_batch`. `read_part` has the same issue on `get()`.

On a 2-core Kind worker there are only a handful of shards, so `buffered(8)` due-eval still serializes unrelated keys.

Put `Arc<RwLock<PartitionState>>` back in the map: clone the `Arc`, drop the DashMap guard, then take the partition lock. No freeze protocol needed.

src/runtime/operators/window/store/backend/inmem.rs
```diff
@@ -318,56 +399,56 @@ impl WindowOperatorStore for InMemWindowStore {
...
     async fn checkpoint(&self, namespace: &StateNamespace) -> Result<WindowBackendSnapshot> {
-        let store = self.state.read().await;
-        let snapshot = store
+        let partitions = self
             .partitions
             .iter()
-            .filter(|(partition, _)| partition.namespace.as_slice() == namespace.bytes.as_slice())
+            .filter(|entry| entry.key().namespace.as_slice() == namespace.bytes.as_slice())
+            .map(|entry| (entry.key().clone(), entry.value().read().clone()))
+            .collect::<HashMap<_, _>>();
```
anovv, line 440: This clones partitions one-by-one then reads triggers with no freeze. The worker cleaner (`maintain`) prunes triggers then partitions concurrently, so a checkpoint can mix pre- and post-prune keys.

Fine for this process-local backend — please note that in the store/README docs rather than reintroducing namespace gates.

anovv (commented):
**Requested changes** (posted as a comment: GitHub does not allow the PR author to use Request changes.)

Second pass after dropping freeze gates / pack fast-paths: that simplification should stay.

**Required:** do not hold the DashMap shard guard across Arrow concat/take. Clone `Arc<RwLock<PartitionState>>` out of the map, drop the map guard, then lock. Same-shard keys currently block for the whole rebuild/select, which reintroduces the #255 contention this PR is meant to remove (just sharded).

Checkpoint vs `maintain` tearing is acceptable for in-mem if the README/store docs say so — do not bring namespace freeze back.

Nits (optional): `PROCESS_DUE_CONCURRENCY` belongs next to `process_due` (eval does not hold partition locks; the cap is “don’t run 256 advance_key at once”). The concurrent-write test only shows completion, not that locks are unshared.

anovv:
**Tests:** drop two, keep packing coverage on the contracts that can actually break.

**Drop**
- `concurrent_partition_writes_complete` — 64 writes finishing does not show unshared locks. No unique assertion.
- `commit_events_packs_rows_into_one_batch_per_key` — two-commit concat is the only extra signal; fold it into `raw_ranges_are_half_open_globally_ordered_and_deduplicated` (second `commit_events`, then the overlapping load). One test then covers unsorted ingest, merge, half-open overlapping runs, and “one packed batch.”

**Keep `assert_packed` only where the shape can break**
- `raw_ranges_…` (select / pack)
- `maintain_prunes_triggers_raw_and_tiles` (prune rebuild)
- `checkpoint_restores_complete_namespace_into_fresh_store` (IPC round-trip)

Remove it from `commit_events_stores_raw_tiles_and_meta_coherently` and `load_window_data_returns_combined_raw_and_tile_view` — those already check cursors/tiles.

Leave the rest (empty/missing, tiles, meta-only write, coherent commit, due paging, namespace-scoped restore, version-checkpoint reject). Each is a different contract and cheap.

anovv:
Addressed the test comment: dropped `concurrent_partition_writes_complete` and `commit_events_packs_rows_into_one_batch_per_key`. The range test now does a second `commit_events` then the overlapping load. `assert_packed` remains on that test, prune, and checkpoint restore only.

Issue #121, opened 2026-03-17
Pipeline rescaling and state redistribution


anovv:
We need to figure out how to properly change parallelism of a running pipeline and how it would affect state storage design

PR #272, opened 2026-08-27
Introduce KeyGroups as routing primitive

### LINKED ISSUE #159: Introduce KeyGroups as routing primitive

anovv:
## Summary
- Adds KeyGroups as the keyed routing primitive (`hash → key_group → subtask`) with job-wide `max_parallelism` (default `= parallelism`, so dests are unchanged vs `hash % p`).
- KeyBy packs by that subtask; `HashPartition` still reads `volga.target_subtask`. Collector dest `i` maps to target `task_index` `i` (numeric sort, so `p ≥ 10` is not lex-broken).
- Rescale / restore is **not** in this PR ([#121](https://github.com/volga-project/volga/issues/121)).

Fixes [#159](https://github.com/volga-project/volga/issues/159).

## Test plan
- [ ] `cargo test --lib key_group`
- [ ] `cargo test --lib pack_by_dest dest_10_routes to_execution_graph_propagates datafusion_key_by`
- [ ] Default `max_p = p`: dests match old `hash % p`
- [ ] Explicit `max_p > p`: same hash keeps the same key group
- [ ] Collector: dest 10 → task 10 when `p ≥ 11`
- [ ] Optional: `cargo run --bin generate-kube-pipeline-schema` if the hand-edited kube schema should be regenerated from `KubePipelineSpec`

### DIFF
--- a/kubevolga/internal/controller/kube_pipeline_spec.schema.json
+++ b/kubevolga/internal/controller/kube_pipeline_spec.schema.json
@@ -39,6 +39,15 @@
         "per_operator": {}
       }
     },
+    "max_parallelism": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "format": "uint",
+      "minimum": 0,
+      "default": null
+    },
     "parallelism": {
       "type": "integer",
       "format": "uint",
--- a/src/api/compiler.rs
+++ b/src/api/compiler.rs
@@ -78,6 +78,7 @@ fn merge_connector_configs(
 fn compile_logical_graph_from_parts(
     sql: &str,
     parallelism: usize,
+    max_parallelism: usize,
     execution_mode: ExecutionMode,
     operator_overrides: &OperatorOverrides,
     event_time: &EventTimeSpec,
@@ -87,6 +88,7 @@ fn compile_logical_graph_from_parts(
     let mut planner = Planner::new(
         PlanningContext::new(df_ctx)
             .with_parallelism(parallelism)
+            .with_max_parallelism(max_parallelism)
             .with_execution_mode(execution_mode),
     );
 
@@ -152,6 +154,7 @@ pub fn compile_logical_graph(spec: &PipelineSpec, connector_overrides: Option<&C
     compile_logical_graph_from_parts(
         sql,
         spec.parallelism,
+        spec.resolved_max_parallelism(),
         spec.execution_mode,
         &spec.operator_overrides,
         &spec.event_time,
--- a/src/api/compiler.rs
+++ b/src/api/compiler.rs
@@ -78,7 +78,6 @@ fn merge_connector_configs(
 fn compile_logical_graph_from_parts(
     sql: &str,
     parallelism: usize,
-    max_parallelism: usize,
     execution_mode: ExecutionMode,
     operator_overrides: &OperatorOverrides,
     event_time: &EventTimeSpec,
@@ -88,7 +87,6 @@ fn compile_logical_graph_from_parts(
     let mut planner = Planner::new(
         PlanningContext::new(df_ctx)
             .with_parallelism(parallelism)
-            .with_max_parallelism(max_parallelism)
             .with_execution_mode(execution_mode),
     );
 
@@ -151,15 +149,22 @@ pub fn compile_logical_graph(spec: &PipelineSpec, connector_overrides: Option<&C
         );
         spec_connector_configs
     };
-    compile_logical_graph_from_parts(
+    let mut graph = compile_logical_graph_from_parts(
         sql,
         spec.parallelism,
-        spec.resolved_max_parallelism(),
         spec.execution_mode,
         &spec.operator_overrides,
         &spec.event_time,
         &connector_configs_owned,
-    )
+    );
+    let max_p = spec.resolved_max_parallelism();
+    assert!(
+        max_p >= spec.parallelism.max(1),
+        "max_parallelism ({max_p}) must be >= parallelism ({})",
+        spec.parallelism
+    );
+    graph.set_max_parallelism(max_p);
+    graph
 }
 
 #[cfg(test)]
--- a/src/api/compiler.rs
+++ b/src/api/compiler.rs
@@ -237,6 +237,7 @@ mod tests {
             .build();
 
         let mut graph1 = compile_logical_graph(&spec, Some(&connector_configs)).to_execution_graph();
+        assert_eq!(graph1.max_parallelism(), spec.resolved_max_parallelism());
         let mut graph2 = compile_logical_graph(&spec, Some(&connector_configs)).to_execution_graph();
 
         let mapping1 = build_mock_vertex_mapping(&graph1);
--- a/src/api/logical_graph.rs
+++ b/src/api/logical_graph.rs
@@ -112,18 +112,12 @@ impl LogicalGraph {
     }
 
     pub fn set_max_parallelism(&mut self, max_parallelism: usize) {
-        self.max_parallelism = max_parallelism.max(1);
+        assert!(max_parallelism >= 1, "max_parallelism must be >= 1");
+        self.max_parallelism = max_parallelism;
     }
 
-    /// Key-group space for this graph, at least every node's parallelism.
     pub fn max_parallelism(&self) -> usize {
-        let node_max = self
-            .graph
-            .node_weights()
-            .map(|n| n.parallelism)
-            .max()
-            .unwrap_or(1);
-        self.max_parallelism.max(node_max).max(1)
+        self.max_parallelism
     }
 
     pub(crate) fn watermark_assign_config_for_column(&self, column_name: String) -> WatermarkAssignConfig {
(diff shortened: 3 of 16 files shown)


anovv:
## Review — please address before merge

Focus: correctness, less plumbing, high-signal tests only.

### Correctness

1. **`LogicalGraph::max_parallelism()` must not silently bump to max node `p`.** That breaks the compatibility contract (`default max_p = p` → dests == `hash % p`). Mixed-`p` graphs (e.g. 2/3/1/1) stamp `max_p = 3` on every vertex, so a KeyBy with `p = 2` routes as `(hash % 3) * 2 / 3` instead of `hash % 2`. Store the resolved value and **fail** if any node exceeds it (`p ≤ max_p` at compile, as in #159). `PipelineSpec::validate` already rejects `max_p < spec.parallelism`; the getter is a second, silent policy.

2. **`max_parallelism as i32` can drop the value.** Spec is `usize`; vertex/context are `i32`. Values `> i32::MAX` wrap negative, then `.max(parallelism).max(1)` falls back to `p` and silently rehashes. Keep `usize` end-to-end, or reject anything above a cap (Flink uses `≤ 32768`). Same for `key_group_of` truncating to `u32` when `max_p > u32::MAX`.

3. **Collector sort must not treat a failed parse as `None` (sorts before task 0).** One bad vertex id shifts every dest. Panic/`expect` — dest `i` must be task `i`. Numeric sort itself is the right `p ≥ 10` fix.

### Simplification

`max_p` is job-wide but is stored and re-clamped in spec, planner, logical graph, every vertex, and runtime context (`max(p).max(1)` ~10 times).

Put it **once** on `ExecutionGraph` (#159 allows graph *or* vertex). `RuntimeContext::max_parallelism()` can read the graph it already holds and fall back to `self.parallelism` when tests construct a context with no graph. That drops `ExecutionVertex.max_parallelism`, `RuntimeContext.with_max_parallelism`, the `tasks.rs` line, and the per-vertex mutation in `to_execution_graph`.

Also drop:
- `PlanningContext::resolved_max_parallelism` — `with_max_parallelism` already clamps
- `with_parallelism` bumping `max_parallelism` as a side effect — resolve once at `Planner::new` / compile
- `key_group_range` (+ `KeyGroupId` re-exports) until #121 — nothing in this PR calls it
- public `Channel::target_task_index` / `vertex_task_index` — a private helper in `collector.rs` is enough
- extra clamps in `pack_by_dest` / `KeyByOperator::open` / `subtask_for_hash` if the primitive asserts `max_p >= p >= 1`

### Tests

**Keep**
- `max_p_equals_p_matches_hash_mod_p` — compatibility contract
- `max_p_gt_p_keeps_key_group_stable` — reason this PR exists
- `dest_10_routes_to_task_10_not_lex_channel` — real `p ≥ 10` bug; highest-signal test in the diff
- one wiring assert that graph/context actually received `max_p`

**Drop / shrink**
- `range_is_inverse_of_subtask` **and** `ranges_cover_all_groups_without_overlap` — same fact; both go if `key_group_range` is deferred
- `pack_by_dest_uses_key_groups_when_max_p_gt_p` (~45 lines of Arrow downcasts) — restates `subtask_for_hash(0/64, 2, 128)`. Existing empty-bucket test already covers packing; if you want a pack-level check, assert extras `{"0","1"}` and stop
- `to_execution_graph_propagates_max_parallelism` **plus** the `max_p == 3` bolt-on — two wiring tests; keep one. The `== 3` case encodes the silent node-max bump to remove

Net: three primitive/routing tests + one plumbing assert.

anovv:
Addressed in 4584ab5:

- `max_p` is stored once on `ExecutionGraph`. `to_execution_graph` fails if any node `p` exceeds it (no silent bump to max node p). Mixed-p tests set `max_p` explicitly.
- `usize` end-to-end on the graph; `RuntimeContext::max_parallelism()` reads the graph and falls back to vertex `p` when tests have no graph.
- Collector panics if the target vertex id is not `_{task_index}`.
- Dropped per-vertex/context copies, `key_group_range` / `KeyGroupId`, public channel helpers, and the extra clamps/tests called out in the review.

anovv:
Follow-up on 4584ab5: the previous items look fixed (graph-level `max_p`, fail-loud collector parse, no node-max bump, tests cut). Remaining:

**`Planner::new` still silently raises `max_p`:**

```rust
logical_graph.set_max_parallelism(context.max_parallelism.max(context.parallelism).max(1));
```

`PlanningContext` defaults `max_parallelism` to `1`, so `with_parallelism(8)` becomes `max_p = 8` via `max()`. That is the intended default, but `with_max_parallelism(2).with_parallelism(8)` is also swallowed instead of failing. Same class of bug as the node-max bump, one layer up. Compile path is safe (`PipelineSpec::validate`), direct `Planner` construction is not.

**Drop `PlanningContext.max_parallelism`.** Default job `max_p` is just `p`:

1. `Planner::new` → `set_max_parallelism(context.parallelism)`
2. Compiler overwrites with `spec.resolved_max_parallelism()` after `Planner::new` (assert `>= p`, do not `max()`)
3. Delete `PlanningContext::{max_parallelism, with_max_parallelism}` and the extra `compile_logical_graph_from_parts` argument if the compiler can set it on the planner

Also unused:
- `pub use key_group::{key_group_of, subtask_for_hash, subtask_of}` in `common/mod.rs` — `pack.rs` imports the module path
- `LogicalGraph::max_parallelism()` getter — `to_execution_graph` already reads the field

anovv:
Addressed in the latest commit:

- `Planner::new` sets `max_p = parallelism` only. `with_max_parallelism(2).with_parallelism(8)` can no longer be swallowed.
- Dropped `PlanningContext::{max_parallelism, with_max_parallelism}` and the extra `compile_logical_graph_from_parts` arg. Compiler overwrites via `spec.resolved_max_parallelism()` with `assert!(max_p >= p)` (no `max()`).
- Removed unused `pub use` of key-group fns and `LogicalGraph::max_parallelism()`.

PR #276, opened 2026-08-29
Remove unused operator chaining runtime

anovv:
## Summary
- Delete `ChainedOperator` (mpsc hops, panicking `poll_next`/`checkpoint`) and `OperatorConfig::ChainedConfig` / `OperatorType::ChainedSourceSink`. Nothing in production enabled chaining (`from_linear_operators(..., chained)` was always `false`).
- Keep `group_operators_for_chaining` as a planner helper that returns `Vec<Vec<OperatorConfig>>` (split on Hash). Not wired into execution; for a later in-process chaining PR.
- StreamTask / worker checkpoint inject now treat only `OperatorType::Source` as sources.

This is PR 1 of 2 for the StreamTask redesign (cleanup so the new loop never special-cases chains). Tests not run yet; will follow after the series is up for review.

## Test plan
- [ ] `./scripts/test unit` (after this PR is reviewed / series is in place)
- [ ] Confirm a linear graph (`from_linear_operators`) still has one vertex per operator
- [ ] `group_operators_for_chaining_keeps_forward_ops_together` and `group_operators_for_chaining_splits_at_key_by_hash`


src/api/logical_graph.rs
```diff
@@ -611,6 +621,23 @@ mod tests {
     use datafusion::prelude::{col, lit};
     use std::sync::Arc;
 
+    fn filter_op() -> OperatorConfig {
```
anovv, line 624: delete this and group_operators_for_chaining_keeps_forward_ops_together

PR #277, opened 2026-08-29
Make StreamTask a dispatcher with push-style operators

anovv:
## Summary
- Stacked on #276. Replaces `poll_next` / `set_input` with a task-owned mailbox: processors implement `process_data` / `handle_watermark` / `handle_barrier`, sources implement `fetch_next`.
- Run loop lives in `src/runtime/stream_task/` (`processor_loop` + `source_loop`). The inbound stream is the mailbox. `InputProgress` owns `WatermarkManager` + `CheckpointAligner` (assign/merge/timer/align). Assigned watermarks emit after the drained data batch.
- `Operator` is `Source(Box<dyn SourceOperator>)` | `Stream(Box<dyn StreamOperator>)`. Map/Join/Sink/Aggregate use `HasOperatorBase` + a blanket `OperatorTrait`; KeyBy/Window/WindowRequest/Source keep explicit lifecycle.
- Shared `TaskCtx` with one WM emit helper. Ingest stamp and `Finished` on MAX watermark are set in the loops, not `TransportOutput`. Lag gauge is emit-only. Oneshot run/close share `wait_for_oneshot`. Sink does not forward data.
- `fetch_next()` has no interrupt argument (handle stays inside the source). Sources must not return watermarks.

## Test plan
- [x] `./scripts/test default` locally
- [ ] After #276 is merged, retarget this PR to `master` if GitHub does not do it automatically
- [ ] Confirm source checkpoint inject, MAX watermark finish, and window ingest-then-watermark still behave as before


anovv:
## Review

The cutover matches the design: task-owned mailbox, `process_data` / `handle_watermark` / `handle_barrier` / `fetch_next`, one `TransportOutput` emit path, assign/merge/align on the task.

### What holds up

The split is the right one. `processor_loop` and `source_loop` own wait points; `MailboxFront` owns assign/merge/align; operators only react. Window ingest-then-watermark is preserved (and is better when assign is on the same vertex: the whole ready data batch is ingested before assigned WMs fire). Barrier path still flushes pending WMs, snapshots, then forwards. Sink no longer forwards data and flushes on the barrier. Tests using `VecOutput` are the right contract.

`Join` dropping `left_buffer` / `right_buffer` is fine — those were written and never read; it was already a pass-through.

### Issues

**1. Upstream watermarks lose extras and trace**

Old preprocess cloned the watermark and only updated value + `upstream_vertex_id`. `MailboxFront::on_upstream_watermark` builds a fresh `WatermarkMessage::new(...)`, which allocates empty metadata.

```rust
if let Some(new_watermark) = self.manager.merge_watermark(watermark.clone()).await {
    let stamp = watermark
        .metadata
        .ingest_timestamp
        .unwrap_or_else(StreamTask::now_ms);
    out.push(self.stamped_watermark(new_watermark, stamp));
}
```

With `MESSAGE_TRACE_ENABLED`, the hop list is reset to this vertex only. Same for any extras on the WM. Mutate the incoming message (or copy `metadata.extras` / trace) instead of constructing a new one.

**2. Source idle is now counted — that is a metrics change**

Old processors wrapped the mailbox in `input_with_idle_tracking`. Sources never did; `fetch_next` wait was residual busy.

`source_loop` now records idle around every `fetch_next` / timer wait. Source busy/idle gauges will move (idle up, busy down) on any blocking fetch (Kafka poll, datagen sleep). The design doc’s acceptance said those gauges stay the same. Either treat this as an intentional fix and say so, or keep source idle uncounted until you mean to change the dashboards.

**3. Source checkpoint still cannot interrupt `fetch_next` via the run loop**

The design had `fetch_next + WM timer + checkpoint_rx` in one `select!`. The loop still `try_recv`s at the top, then blocks in `fetch_next(None)`.

That matches today’s worker (cancel handle, then send the trigger), but the race is now more visible: cancel can yield `Interrupted` before the oneshot/mpsc trigger is in the channel. The next line treats `Interrupted` as a no-op and starts another fetch. A long Kafka/datagen wait then delays the barrier until that fetch ends.

Worth selecting on `checkpoint_receiver` next to `fetch_next` (and passing the handle interrupt), so an `Interrupted` with an empty channel does not start another blocking pull.

**4. `wait_for_run` / `wait_for_close` treat a dropped oneshot as success**

```rust
match tokio::time::timeout(timeout, receiver).await {
    Ok(_) => {}
    Err(_) => panic!(...),
}
```

`Ok(Err(RecvError))` — sender dropped — proceeds as if the signal arrived. Watch had the same shape, but oneshot fails immediately on drop. If `StreamTask` is dropped mid-wait, the run loop starts (or closes) without a real signal. Match `Ok(Ok(()))` only.

### Smaller notes

- `INGEST_MAX_RECORDS` (64k) now applies to every processor, not just window. Map/KeyBy will ingest a ready burst in one `process_data`. That matches the doc; just be aware output bursts to downstream get larger.
- `MailboxFront` tests cover assign/merge/flush-before-barrier, not the loop’s “drain batch, then emit assigned WMs” order. A small `processor_loop` test (or a fake operator that records call order) would lock the thing the PR is selling.
- Source `Idle` → MAX WM → `TransportOutput` sets `Finished` still looks correct. Processor mailbox EOF still finishes without a downstream MAX (same as before).

Stacked on #276 as noted; retarget after that merges. I would not merge this until (1) is fixed; (2) and (3) are either fixed or called out as accepted.

anovv:
## Follow-up: simplify (less code / simpler logic)

The dispatcher model is right. Most of the leftover weight is **three copies of the same fields**, a **MailboxFront that only forwards**, and **operator boilerplate** that the new traits were supposed to shrink.

### Highest leverage

**1. `enum Operator { Source, Stream }` as the design said**

You still have a fat enum plus three identical match arms (`as_stream_mut`, `as_trait_mut`, `as_trait`) plus `OperatorTrait for Operator`. That is the old `poll_next` dispatch with new names.

```rust
enum Operator {
    Source(Box<dyn SourceOperator>),
    Stream(Box<dyn StreamOperator>),
}
```

`run` already branches on source vs not. `create_operator` becomes the only match. Deletes `as_*` and the wrapper trait impl.

**2. Stop implementing `OperatorTrait` on every map/join/keyby/sink**

Each of those is the same four forwards to `OperatorBase`. After (1), lifecycle can live on `OperatorBase` (or default trait methods), and Map/KeyBy only implement `process_data`. That is most of the operator-file churn in this PR.

**3. Drop `MailboxFront` as a type, or shrink it to two methods**

It is `WatermarkManager` + `CheckpointAligner` plus `merge_emits` on every path. `assign_on_data` / `assign_and_merge` / `merge_assigned` / `merge_emits` / `stamped_watermark` are one idea.

Keep the manager and aligner as fields on the processor loop. You need:

- `on_data` → optional assigned+merged WM
- `on_upstream_wm` / `on_timer` / `flush` → `Vec<WM>`
- `on_barrier` → align + flush

That is the mailbox invariant. A wrapper struct is optional.

### Same logic, less code in the loops

**4. One emit helper**

`emit_watermarks` (processor, via `handle_watermark`) and `emit_assigned` (source, via `TransportOutput`) are the same loop. Difference is “ask the operator” vs “emit raw”. One function with a `FnMut` or a `bool through_op` is enough.

**5. `stamp_ingest_if_missing` off `TransportOutput`**

Only sources stamp. Stamp in `source_loop` before `emit`, and let `TransportOutput` always mean “send + metrics + upstream id”. Removes the flag from every `ctx.output(...)`.

**6. `Finished` on MAX out of `TransportOutput`**

Hidden side effect. Source Idle and processor mailbox EOF already set `Finished` in the loop. Do the same when a MAX WM is about to be emitted, and `TransportOutput` is just I/O.

**7. Lag gauge is set twice**

`MailboxFront::stamped_watermark` and `TransportOutput::emit` both call `set_watermark_lag_gauge`. Keep it on emit only.

**8. Merge `wait_for_run` / `wait_for_close`**

Same 50s oneshot wait. One `wait_for(receiver)` (and only treat `Ok(())` as success, from the earlier review).

**9. `RunParams` is `StreamTask` copied field-by-field**

`start()` is ~50 lines of clone-into-struct. Group signals (`run` / `close` / `checkpoint`) and watermark Arcs so the call is a handful of groups, or pass a smaller owned bag. You cannot move `StreamTask` itself (the actor keeps it).

### Dead or test-only surface

**10. `drain_ready_inputs` + `NextInputs`**

`processor_loop` only calls `drain_ready_after`. `NextInputs` exists for `operator_test.rs`. Point those tests at `drain_ready_after` (or a tiny test helper) and delete the enum.

**11. `fetch_next(interrupt: Option<&SourceInterrupt>)`**

The loop always passes `None`; the source uses the handle interrupt anyway. Drop the argument.

**12. Source `merge_watermark` on a fetched `Message::Watermark`**

Sources almost always return `Regular`. Assigned WMs already go through `on_data_message` + `emit_assigned`. If nothing real emits a WM from `fetch_next`, delete that branch.

**13. `record_recv` cloning control messages**

```rust
record_recv(&ctx, &Message::Watermark(wm.clone()));
```

`record_metrics` no-ops on control. Call `record_control_propagation` on `&wm` / `&barrier` and skip the clone.

### What I would not simplify yet

- **File split** (`run` / `processor` / `source` / `mailbox` / `checkpoint`). The loops are readable this way. Merging `ctx.rs` into `output.rs` is optional, not important.
- **`TaskCtx`**. It is the thing that stopped the loops from taking 12 parameters. Keep it.
- **`dyn Output`**. Needed for `VecOutput` in tests and later `ChainedOutput`.
- **Join as pass-through**. Already as small as it can be until join is real.

### If you only do a few

In this PR: **(8) wait helper, (7) one lag gauge, (10) drop `NextInputs`, (5) stamp only on source, (13) stop cloning WMs**. Small and local.

Next PR or a follow-up commit: **(1) + (2)** — that is the real line-count win, and it matches the doc. **(3)** if `MailboxFront` still feels like a second event loop after you stare at it.

anovv:
Landed the remaining review items on this PR (no follow-up).

**First review**
1. Incoming upstream watermarks keep extras/trace; only value + vertex id are rewritten (`mailbox_front_preserves_upstream_watermark_extras`).
2. Idle stays on processor mailbox wait only; source `fetch_next` is not counted as idle.
3. Source loop `select!`s checkpoint + fetch + WM timer. On `Interrupted`, if the source is not stopped, it `recv()`s the checkpoint id instead of starting another blocking fetch.
4. Oneshot success is only `Ok(Ok(()))`; a dropped sender panics. Run/close share `wait_for_oneshot`.

**Second review (all here)**
1. `Operator` is now `Source(Box<dyn SourceOperator>)` | `Stream(Box<dyn StreamOperator>)`. `create_operator` is the only fat match; `as_stream_mut` / `as_trait*` are gone.
2. Map/Join/Sink/Aggregate use `HasOperatorBase` + a blanket `OperatorTrait` (no lifecycle macros). KeyBy/Window/WindowRequest/Source keep explicit open/close/checkpoint.
3. Dropped `MailboxFront`. Processor holds `WatermarkManager` + `CheckpointAligner` and calls mailbox helpers. Tests drive the same helpers.
4. One emit helper (`TaskCtx::emit_watermarks`) for through-op vs raw.
5. Ingest stamp only on the source path.
6. `Finished` on MAX watermark is set in the loops, not `TransportOutput`.
7. Lag gauge only on `TransportOutput::emit`.
8. Merged `wait_for_run` / `wait_for_close` into `wait_for_oneshot`.
9. `RunParams` groups signals + watermark Arcs.
10. Deleted `NextInputs` / `drain_ready_inputs`; tests use `drain_ready_after`.
11. `fetch_next()` no longer takes `interrupt`; the source still uses `SourceFunction::fetch(interrupt)` internally.
12. Source fetch must not return watermarks (panic if it does); assign/merge stays in the task.
13. Recv metrics wrap/unwrap control messages instead of cloning them.

Left as-is per review: file split, `TaskCtx`, Join as pass-through.

anovv:
Kube CI failure on `test_kube_multi_worker_window_checkpoint_restore` was a source-loop deadlock, not a flake.

The review cleanup put checkpoint wait in the biased `select!` and then **blocked** on `checkpoint_receiver.recv()` after `FetchResult::Interrupted`. Worker order is `cancel()` then send the id, so the leftover latch often fired after the id was already consumed. That source hung; window watermark min-merge stalled; the upsert sink froze (timed out waiting to advance past 480) before the kill.

`d33be99` restores the previous contract: `try_recv()` the barrier at the top of the loop, treat `Interrupted` as a no-op. Same as master / early in this PR.

Issue #261, opened 2026-08-25
Umbrella: streaming throughput after TCP (Kind unlimited / Window path)


anovv:
## Why this exists

Kind unlimited soaks after per-edge TCP ([#254](https://github.com/volga-project/volga/pull/254)) still sit around **~11k source rec/s** (16384 keys, RANGE 10ms, SUM/COUNT/AVG, `p=4`, 2 workers). Workers are ~1.0 CPU of a 2-core limit — not Kind throttle. This issue is the map so we do not lose the bottleneck list in individual tickets.

Repro: `kubevolga/hack/bench/unlimited.yaml` + `volga-bench` (`rate: null`). Grafana: source vs sink throughput (not sum of hops). Dump extras: WO ingest/`process_due`, key skew.

**Do not pile these onto transport PRs.** One concern per PR; re-measure soak after each merge.

## What the soak actually showed

| Layer | What we saw |
|---|---|
| Transport | TCP is no longer the silent stall. Still **flush every frame** ([#256](https://github.com/volga-project/volga/issues/256)). |
| Watermarks | Assigner emits on **every advancing source batch**. OOO does not cap fire rate. Window `process_due` runs at ingest-batch rate ([#260](https://github.com/volga-project/volga/issues/260)). |
| Window | In-mem: **one RecordBatch per row**, **one `RwLock` over all keys**, `process_due` `join_all` uncapped ([#257](https://github.com/volga-project/volga/issues/257), [#255](https://github.com/volga-project/volga/issues/255)). RANGE 10ms + 1s/5s tiling is now skipped when window &lt; finest granule (local tile change; keep it). |
| Shuffle | KeyBy **splits a batch into one `KeyedMessage` per key** ([#263](https://github.com/volga-project/volga/issues/263)). Hash **`Key::new` recomputes** ([#258](https://github.com/volga-project/volga/issues/258)). Source→KeyBy is RoundRobin; Forward still TODO ([#142](https://github.com/volga-project/volga/issues/142)). Low key counts hash-skew onto Window tasks. |
| Runtime | Serial `poll_next` per task is correct. Extra tokio threads do not help a global store lock ([#153](https://github.com/volga-project/volga/issues/153)). Idle time metric wraps **entire** `poll_next` (ingest + `process_due` mislabeled idle) ([#262](https://github.com/volga-project/volga/issues/262)). |

## Suggested order

1. **Measure Window in isolation** so kube e2e is not the only ranking tool.
[...]

Issue #167, opened 2026-07-31
Store large operator checkpoints outside the control plane


anovv:
## Problem

`InMemWindowStore` currently serializes an entire namespace into `WindowBackendSnapshot::InMemory`. That payload is sent from the worker to the master in checkpoint reporting, retained by the master's checkpoint store, and sent back through worker configuration during restore.

This does not scale with state size: it can exceed gRPC message limits, creates several full-payload memory copies, increases checkpoint pauses, and turns the master into a state-data bottleneck. Inline in-memory checkpoints are now explicitly development/test-only and capped at 3 MiB.

## Proposed direction

Introduce worker-accessible checkpoint blob storage, separate from the master's completed-checkpoint metadata store:

1. The worker serializes and uploads an immutable operator checkpoint blob before reporting checkpoint completion.
2. The operator checkpoint contains a small reference with object identity, size, checksum, and format/version instead of the state bytes.
3. The master stores and forwards the opaque reference through the existing `SerializedCheckpoint` / `SerializedRestore` flow.
4. A replacement worker downloads and validates the blob directly during operator restore.
5. Checkpoint retention removes blobs that are no longer referenced by retained completed checkpoints.

Versioned backends such as the proposed Scylla window store should continue returning small backend-version metadata and should not upload full state.

## Requirements

- Configure a checkpoint blob store accessible to workers; object storage should be the production path.
- Make uploads immutable, idempotent, retryable, and scoped by pipeline/checkpoint/task/operator identity.
- Validate payload size and checksum before restore.
- Define cleanup behavior for failed uploads, incomplete checkpoints, and expired completed checkpoints.
- Keep bulk state bytes out of master/worker control-plane RPCs.
- Preserve the current opaque operator checkpoint and master restore-planning interfaces.

## Tests

- Restore a checkpoint larger than the control-plane gRPC limit on a different worker.
- Verify corrupted or incomplete blobs are rejected.
- Verify retrying checkpoint upload/reporting is idempotent.
- Verify checkpoint retention cleans up unreferenced blobs without deleting retained checkpoints.

PR #278, opened 2026-08-29
Emit watermarks on event time (plus wall cap); yield process_due

### LINKED ISSUE #273: Watermark emit should be event-time (plus wall-clock cap); 200ms processing-time made process_due stall
## Why

Kind unlimited soak after [#267](https://github.com/volga-project/volga/pull/267) (200ms **processing-time** `emit_interval`) did **not** OOM. `process_due` p99 **hit the 5s histogram cap**, Window rx sat ~8180 (full), Source/KeyBy were backpressured. In-mem state grew **379k → 2.8M rows** / **~1.2 GB** because prune/eval could not keep up.

Same cell with `emit_interval_ms: 0`: state **~80 MB**, `process_due` p99 **~21 ms**.

Not a watermark-correctness bug (`WM = max_seen − OOO`). One **processing-time** interval let a huge **event-time** jump land in a single exclusive `handle_control` → `process_due`. Channel backpressure bounds **in-flight records** (8192), not **store size** (`W − RANGE` + lag while `W` is stale).

Child of [#261](https://github.com/volga-project/volga/issues/261). One concern per PR; re-soak. Do not pile onto unrelated PRs.

## What the 200ms cell actually was

Soak: `SUM/COUNT/AVG/MIN/MAX(...) OVER (... RANGE 10s)`, 16384 keys, IncrementalTimestamp `step_ms=1` per key, unlimited datagen, `INGEST_MAX_RECORDS = 64k` per poll.

1. WO is two-phase by design: ingest appends raw/tiles and enqueues `RowEmit`; emit/eval/prune run on WM in `handle_control` (exclusive with ingest). Keep this. Do **not** fold OVER emit into `insert_batch` (event-time row is not complete until `WM ≥ T`; Scylla selective I/O in [#157](https://github.com/volga-project/volga/pull/157) is append-on-ingest + coverage-on-advance). Hop later is `WindowEnd` on the same flow.
2. Between WMs the task loops ingest (up to 64k rows per poll until a peeked control message). All those rows become due work. **BP is off during this stretch** — only on while `process_due` runs.
3. 200ms **wall-clock** emit + flying event time ⇒ one WM can cover seconds of event time. `process_due` walks that due-set for seconds. Queues fill; if the next WM cannot be enqueued (or loses the post-BP send race), the next gulp is queue-sized, not interval-sized → lag grows → store grows → worse next wave.
4. `INGEST_MAX=64k` made polls fat; it was **not** the 5s. The 5s is due-set size. More `process_due` parallelism ([#270](https://github.com/volga-project/volga/pull/270)) does not fix a pathological due-set.

**Equilibrium** (when it works): applied `W` keeps up with ingested max event time → prune floor `W − max_wl` → state ≈ RANGE + small lag. Async `maintain` (1s tick) is fine; unbounded growth is **stale `W`**, not “cleaner too slow.” Fired triggers lingering until maintain can make `process_due` heavier but are not the primary spiral.

**`emit_interval` is a due-set size knob (in event time), not a “make processing faster” knob.** Smaller interval = smaller chunks, more often (U-curve: too small → control overhead; too large → fat due-sets). Users should not need to tune it for stability once (1)+(2) land.

## Arroyo (reference)

Same ingest-then-WM shape; **no** reserved control slot. [`BatchSender`](https://github.com/ArroyoSystems/arroyo/blob/f9eacea1/crates/arroyo-operator/src/context.rs) counts **rows** toward 8192; a `Signal` (WM/barrier) also costs 1 — it **blocks** when full like data.

Their [`ExpressionWatermark` / `period_micros`](https://github.com/ArroyoSystems/arroyo/blob/f9eacea1/crates/arroyo-worker/src/arrow/watermark_generator.rs) is **event-time**: `collect(batch)` then `broadcast_watermark` when event time since last emit exceeds the period. Downstream handles **one mailbox message at a time** (no 64k gulp between WMs). Idle partitions use `Watermark::Idle` (merge exclusion), separate from periodic emit. Tumble/hop keep aggregators — still OOM on long buffer + slow WM without due-set caps.

Volga target after (1)+(2): ET period caps the due-set; yield avoids one WM freezing the task for seconds — closer to Arroyo’s interleaving without requiring a reserved FIFO slot first.

### 1. Event-time emit interval + wall cap (one knob, two clocks)

`is_due` today is `now − last_emit_at` (processing time only) — wrong for flying event time (`step_ms=1`).

**Emit when:**
- **Event time** has moved by `emit_interval_ms` since last emit (primary while data flows), **or**
- **Wall clock** has moved by the **same** `emit_interval_ms` since last emit **and** there is an **unpublished** candidate (`max_seen − OOO > last_emitted`).

Keep [#260](https://github.com/volga-project/volga/issues/260) timer + EOF/barrier `flush`. **Do not spam:** after wall flush, candidate is published → no deadline until the next record advances `max_seen` (same pattern as today’s `next_emit_deadline` / `unpublished` gate). Never had data → no deadline.

**Config:** one user-facing `emit_interval_ms` (due-set period + idle flush latency). **`idle_timeout_ms` stays separate** — it is merge **min** semantics (“drop a silent *upstream* from `active_upstreams`”), not assigner flush. Do **not** tie merge idle to `emit_interval` (200ms merge-idle would drop slow KeyBy tasks as dead; 10s would block frontier on a crashed sibling).

Inject WMs **after the data batch that crossed the ET threshold** (natural FIFO: records already queued stay before the WM; `T ≥` their timestamps).

Until this lands, soak YAML should keep `emit_interval_ms: 0` (or a small ET interval). Do not treat 200ms **processing-time** as safe for this SQL.

### 2. Yield `process_due`

Do not hold exclusive `handle_control` until the entire due-set + concat finishes (seconds → mailbox fills → post-BP data burst before WM).

**Semantics (do not break WM ownership):**
- One WM still owns the work: pages of due-set for **that** `T`.
- **Do not** take the **next** WM or call `next_inputs()` for data while this WM’s due-set has remaining pages.
- When pages are done: forward WM / emit pending as today, then resume normal poll.

**Mechanism — no new `OperatorPollResult`:** use existing `Continue` / `Ready`. Window keeps resume state (`through`, page cursor). Yield after **one `stream_due` page** (256 triggers today — same quantum as `PAGE_SIZE`; optional time budget only if single-key `advance_key` can exceed it). Prefer **`Ready` per page** (downstream + output BP see progress); do not busy-loop on empty `Continue`.

**Metrics:** record **`wo_wm_process_ms` once per WM** (start first page → done last page) so p99 still signals “this WM owned the task for N seconds.” Optional second histogram for page latency; do **not** put per-page samples on the WM histogram (would hide 5s stalls).

### 3. Control can enqueue while data queue is full — try send policy first, reserved slot optional

**Problem:** assigner is **source-only**; WMs travel the same FIFO as data (Source → KeyBy → Window). When the 8192 **record** cap is full, `send(record)` blocks; source may not be in the emit `select!`. After unblock, **biased** `poll_next` often sends **more data** before the due WM → Window’s next prefix is another 8192, not one ET period.

**Not:** jump WM ahead of queued data (would violate ordering / drop rows as late).

**Try first (no transport change):** after `send` unblocks (or at start of each source loop iteration), **flush due WMs before the next record**; fix biased `select!` so timer/unpublished candidate is not starved by immediate datagen. Same FIFO, same ordering — “generation in front” of the *next* send, not reordering the mailbox.

**Optional reserved tail (transport):** count data records toward 8192; allow 1 control slot (coalesce: if tail is already WM, raise `T`). Helps if send-WM-first still loses the race (Window drains data before Source runs). Checkpoint barriers may need the same treatment. **Defer to a follow-up** unless re-soak after (1)+(2) still shows WMs stuck behind full data queues.

Separate WM side channel is out unless paired with strict seq-N apply-after-ingest; prefer reserved FIFO or send policy.

## Explicitly not in (1)+(2)

- **Ingest credit / “stop drain while due remaining”** — mostly redundant once ET-on-data injects WMs frequently; `handle_control` already does not ingest during `process_due`. BP failure mode is **between** WMs + post-BP gulp, not parallel ingest during `process_due`.
- **Prune on hot path** — async `maintain` stays; growth is stale `W`, not cleaner period.
- **Shrinking `DEFAULT_QUEUE_RECORDS` or `INGEST_MAX_RECORDS` as the fix** — shrinks post-stall gulp only; does not cap ingest between WMs when BP is off.

### Operator-agnostic mem pressure

In-mem vs Scylla/Slate does not change the shape: RANGE + rare WMs = unbounded retained state (worker RSS vs cluster disk). Safety net so remote is not a bottomless sink ([#157](https://github.com/volga-project/volga/pull/157) already sketches this).

- Charge at the **store write**, not WO shapes: one `retained_bytes` (or entry count) per namespace in `StateResourceTracker`. `+= payload.len()` on commit, `-=` the same on prune. No per-row size column; at most a bucket/namespace total, plus checkpoint that total (or recompute on restore).
- Local cache occupancy (Foyer) is a **second** gauge. Do not use process RSS as the controller (kill switch only).
- **Queue BP vs mem pressure:** queue full → sender waits; receiver processes FIFO as usual. Over quota → do not admit **more data**; still `recv` in-flight records; then WM; then prune. If several ops refuse `recv`, WMs stuck behind data deadlock — control must still flow (send policy or reserved slot).

### Out of scope here

- OVER emit-on-ingest / a second OVER operator
- Byte-credit transport (Flink-style); `queued_bytes` stays metrics unless we later cap send on bytes
- Speeding `process_due` on a still-huge due-set as the primary fix
- Changing 8192 from records to something else

## Acceptance

First PR(s) for **(1)+(2)**; **(3)** only if re-soak shows WMs/barriers still stuck behind full data queues after send-WM-first:

- Unlimited 180s cell (`kubevolga/hack/bench/unlimited.yaml`, p=4, 16384 keys, RANGE 10s) does not hit 5s `process_due` / GB-scale state growth at a reasonable ET `emit_interval_ms` (e.g. 200ms event-time + same wall cap).
- Compare to emit=0: ~80 MB state, `process_due` p99 ~20 ms (sanity bound, not the product default).
- Quiet source with held candidate advances after one `emit_interval_ms` wall wait; **no** repeated WM spam while idle.
- Merge `idle_timeout_ms` still excludes silent upstreams independently of emit period.
- WMs still do not overtake records on the same channel.

## Related

- `src/runtime/watermark/manager.rs` (`is_due`, `next_emit_deadline`, `unpublished`, `idle_timeout_ms` vs assigner)
- `src/runtime/stream_task.rs` (source `select!`, `send_source_assigned_watermarks`, biased poll)
- `src/runtime/operators/window/operator.rs` (`INGEST_MAX_RECORDS`, `handle_control` → `process_due`, `PAGE_SIZE`)
- `src/transport/batch_channel.rs`
- [#260](https://github.com/volga-project/volga/issues/260) / [#267](https://github.com/volga-project/volga/pull/267) processing-time coalescer (superseded for soak by ET emit)
- [#255](https://github.com/volga-project/volga/issues/255) / [#270](https://github.com/volga-project/volga/pull/270) `process_due` cap / store locks
- [#157](https://github.com/volga-project/volga/pull/157) Scylla store (`StateResourceTracker` scaffold)

anovv:
## Summary
- `emit_interval_ms` is event-time (emit when candidate moved by X) with the same X as a wall-clock cap for a quiet source that still has an unpublished candidate. `idle_timeout_ms` stays merge-only.
- Window `process_due` yields after one `stream_due` page (`Ready`); the same WM owns remaining pages — no `next_inputs()` until the due-set is done. `wo_wm_process_ms` is recorded once per WM.

Send-policy / reserved FIFO slot (#273 item 3) is **not** in this PR.

Fixes #273. Child of #261.

## Test plan
- [ ] `unset RUSTC_WRAPPER && ./scripts/test unit --filter 'test(/watermark::|operators::window::/)'`
- [ ] `./scripts/test unit` (or default) after review
- [ ] Kind soak 180s `unlimited.yaml` p=4, 16384 keys, RANGE 10s, `emit_interval_ms: 200` (now ET): no 5s `process_due` / GB state vs emit=0 baseline (~80 MB, p99 ~20 ms)
- [ ] Quiet source: one wall-cap advance then stop (no idle WM spam)
- [ ] Rebuild `volga:latest` + `kind load` before soak (worker/runtime change)

### DIFF
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -22,5 +22,9 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s"
+  "transport.tcp_connect_retry_delay": "1s",
+  "window.ingest_max_records": 65536,
+  "window.partition_io_concurrency": 32,
+  "window.due_page_size": 256,
+  "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -24,7 +24,7 @@
   "transport.tcp_connect_max_retries": 10,
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
-  "window.partition_io_concurrency": 32,
+  "window.ingest_key_concurrency": 32,
   "window.due_page_size": 256,
   "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.due_page_size": 256,
-  "window.process_due_concurrency": 16
+  "window.process_page_size": 256,
+  "window.process_key_concurrency": 16
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -22,5 +22,9 @@
   "worker.register_connect_timeout": "200ms",
   "worker.register_rpc_timeout": "500ms",
   "transport.tcp_connect_max_retries": 2,
-  "transport.tcp_connect_retry_delay": "50ms"
+  "transport.tcp_connect_retry_delay": "50ms",
+  "window.ingest_max_records": 65536,
+  "window.partition_io_concurrency": 32,
+  "window.due_page_size": 256,
+  "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -24,7 +24,7 @@
   "transport.tcp_connect_max_retries": 2,
   "transport.tcp_connect_retry_delay": "50ms",
   "window.ingest_max_records": 65536,
-  "window.partition_io_concurrency": 32,
+  "window.ingest_key_concurrency": 32,
   "window.due_page_size": 256,
   "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "50ms",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.due_page_size": 256,
-  "window.process_due_concurrency": 16
+  "window.process_page_size": 256,
+  "window.process_key_concurrency": 16
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -22,5 +22,9 @@
   "worker.register_connect_timeout": "3s",
   "worker.register_rpc_timeout": "3s",
   "transport.tcp_connect_max_retries": 10,
-  "transport.tcp_connect_retry_delay": "1s"
+  "transport.tcp_connect_retry_delay": "1s",
+  "window.ingest_max_records": 65536,
+  "window.partition_io_concurrency": 32,
+  "window.due_page_size": 256,
+  "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -24,7 +24,7 @@
   "transport.tcp_connect_max_retries": 10,
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
-  "window.partition_io_concurrency": 32,
+  "window.ingest_key_concurrency": 32,
   "window.due_page_size": 256,
   "window.process_due_concurrency": 16
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.due_page_size": 256,
-  "window.process_due_concurrency": 16
+  "window.process_page_size": 256,
+  "window.process_key_concurrency": 16
 }
--- a/docker/pipeline_spec.demo.json
+++ b/docker/pipeline_spec.demo.json
@@ -30,7 +30,7 @@
           "fields": {
             "value": {
               "Key": {
-                "num_unique": 1
+                "num_unique_keys": 1
               }
             }
           },
--- a/kubevolga/config/samples/volga_v1alpha1_pipeline.yaml
+++ b/kubevolga/config/samples/volga_v1alpha1_pipeline.yaml
@@ -82,7 +82,7 @@ spec:
               "run_for_s": null,
               "batch_size": 5,
               "fields": {
-                "value": { "Key": { "num_unique": 6 } }
+                "value": { "Key": { "num_unique_keys": 6 } }
               },
               "replayable": true
             }
--- a/kubevolga/hack/bench/dashboards/volga-bench.json
+++ b/kubevolga/hack/bench/dashboards/volga-bench.json
@@ -7,7 +7,7 @@
   ],
   "timezone": "browser",
   "schemaVersion": 39,
-  "version": 8,
+  "version": 9,
   "refresh": "5s",
   "time": {
     "from": "now-15m",
@@ -79,9 +79,9 @@
   },
   "panels": [
     {
-      "id": 10,
+      "id": 20,
       "type": "row",
-      "title": "Job",
+      "title": "Flow",
       "gridPos": {
         "h": 1,
         "w": 24,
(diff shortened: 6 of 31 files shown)


anovv:
## Review

Right split for #273 (1)+(2), but **not merge-ready**: in-mem store tests do not compile, and the window harness still assumes one-shot `process_due`.

Assigner change looks correct: emit when event time moved by `emit_interval_ms`, or when wall clock did and a candidate is still unpublished. `idle_timeout_ms` stays merge-only. Window yield is also right: one WM owns the due-set, `Ready` per `stream_due` page, no `next_inputs()` until the stream ends, `wo_wm_process_ms` recorded once.

### Blocker: `DUE_PAGE_SIZE` is not in scope in `inmem` tests

`inmem.rs` tests call `super::DUE_PAGE_SIZE`. From `inmem::tests`, `super` is the `inmem` module, which does not import the constant (it lives on `backend`). Lib tests fail to compile:

```
error[E0425]: cannot find value `DUE_PAGE_SIZE` in module `super`
   --> src/runtime/operators/window/store/backend/inmem.rs:999:20
```

Same at 1067 and 1154. Import it in the `inmem` `use super::{...}` list (or use `super::super::DUE_PAGE_SIZE`).

### Blocker: `watermark_and_output` runs maintain before the frontier moves

Frontier is stored only when the due stream is exhausted (`continue_due` stream-end path). The harness still does one poll, then `run_maintenance()`:

```rust
// src/test_utils/window/harness.rs
let out = match self.op.poll_next().await {
    OperatorPollResult::Ready(Message::Regular(base)) => base.record_batch,
    other => panic!("expected output on watermark, got {:?}", other),
};
self.run_maintenance().await;
```

That first poll is only the first page. `due_advance` is still live and the frontier is still the old value (unset on the first WM). Maintain then no-ops or prunes against the previous `W`.

`drain_passthrough_watermark` is what actually finishes the due-set and stores the new frontier — **after** maintain. Retention tests such as `raw_and_tiles_are_retained_through_wo_floor` will not prune.

`run_wo_scenario` has the same hole: it keeps only that first page. Fine for 9-row matrix fixtures; wrong at 256+ triggers (the new smoke test avoids the harness, which is why it works).

Harness should drain Regular pages until the WM is produced, **then** maintain, and concat pages. Until that lands, almost every WO test that uses `watermark_and_output` is testing mid-advance state.

### Other notes (non-blocking)

- **Checkpoint vs in-flight due.** `checkpoint()` does not complete or record `due_advance`. Safe today because `poll_next` will not `next_inputs()` (so no barrier) while pages remain. Implicit invariant: if anything later puts a barrier in `pending_messages` during due, StreamTask would snapshot a stale frontier and drop the held WM. Worth a comment on `DueAdvance` / `checkpoint`.
- **ET interval is per message, not per row.** `on_data_message` uses `batch_max` and emits at most one WM per batch. A large batch that spans seconds of event time still produces one fat due-set. Matches “inject after the batch that crossed.” Soak `batch_size: 20` is fine; do not treat `emit_interval_ms: 200` as a due-set cap for 64k-row batches.
- **Item 3 correctly deferred.** Source `select!` is still biased toward `poll_next`; if soak still starves WMs behind a full 8192, that is the follow-up.
- **Assigner tests look good.** `event_time_interval_emits_without_waiting_for_wall` plus existing `coalesce_holds_until_interval` / `timer_does_not_emit_when_already_published` cover ET, wall cap, and no idle spam.

Fix the compile error and drain-then-maintain in the harness before relying on the unit filter in the test plan.

anovv:
## Simplification

The yield and ET emit are the right shape. Most of the extra code is compatibility around **how the WM leaves the operator**, not the paging itself.

### 1. Drop `had_page` and empty Regulars

Biggest cut. Completion is two different machines:

```rust
return if due.had_page {
    OperatorPollResult::Ready(Message::Watermark(due.watermark))
} else {
    self.base.pending_messages.push(Message::Watermark(due.watermark));
    OperatorPollResult::Ready(Message::new(
        None,
        RecordBatch::new_empty(self.output_schema.clone()),
        None,
        None,
    ))
};
```

That empty Regular exists so a no-work WM still looks like the old “batch then pending WM” poll. The harness already loops until it sees a WM. Downstream does not need an empty batch.

Always finish due with `Ready(WM)`. Delete `had_page`. Same for the non-due Emit tail in `handle_control` (pending + empty Regular) — just `Ready(WM)`. StateOnly stays `pending + Continue`.

`DueAdvance` shrinks to `{ pages, watermark, started }`. `wm_ts` is `watermark.watermark_value` (plus the `MAX` → `i64::MAX` map) at finish time.

### 2. Flatten `handle_control` into three cases

Right now Emit+advance returns early, then a shared tail updates the frontier and forwards. Clearer and shorter:

- **does not advance** → `Ready(WM)`
- **StateOnly + advance** → store frontier, pending WM, `Continue`
- **Emit + advance** → start `DueAdvance`, `return continue_due()`

No shared empty-batch tail.

### 3. Bake `DUE_PAGE_SIZE` into the backend

`page_size` is on the trait, re-exported, and passed through operator / inmem / io_planning / retention / semantics. Every caller passes `DUE_PAGE_SIZE`.

Hardcode 256 inside `stream_due`. Drop the argument. The inmem paging test still asserts 256 / 34.

### 4. Merge harness `watermark_and_output` + `drain_passthrough_watermark`

Almost every test is:

```
watermark_and_output(w)
drain_passthrough_watermark()
```

`held_watermark` exists only because the first function now consumes the WM while collecting pages. One `advance(wm) -> RecordBatch` that drains pages, consumes the WM, then runs maintain. Delete `held_watermark` and the drain calls. Tests that assert the WM value can return `(batch, wm)` or keep a one-line helper.

The nonempty-filter / “first empty page” concat in `watermark_and_output` also goes away if the operator never emits empty Regulars (item 1).

---

Leave as-is: holding the due **stream** (better than a resume cursor on the operator), `is_due` (already small), calling `continue_due` immediately instead of returning `Continue` first.

I’d do 1+2 in the operator, then 3+4 as mechanical cleanup.

anovv:
Pushed soak follow-ups (`acb51b7`), not the #273 core:

- `TimeGranularity::Milliseconds` so RANGE 1s can use `[100ms, 500ms]` tiles (1s tiles never fit in a 1s RANGE).
- Datagen `num_unique_keys` (serde alias still accepts `num_unique`); bench YAML `step_ms` on unlimited only; finite rate still derives step. Default `unlimited.yaml` OOO is 1.
- Grafana: drop cumulative operator-records pane; Job/checkpoints sit after Window.

Left out: `docs/`, `/tmp` soak YAMLs/dumps. Kind FT on the unlimited cell is still blocked by the 8 MiB inline checkpoint cap (#167).

PR #283, opened 2026-09-03
fix: include the months component in RANGE window length

cheb0:
## Problem

`get_window_length_ms` decodes an `IntervalMonthDayNano` RANGE bound by summing only `nanoseconds` and `days`, and never reads `months`:

```rust
ScalarValue::IntervalMonthDayNano(Some(v)) => {
    (v.nanoseconds / 1_000_000) + (v.days as i64 * 24 * 60 * 60 * 1000)
}
```

DataFusion normalizes `MONTH` and `YEAR` intervals into the `months` field with `days == 0` and `nanoseconds == 0`, so any month-based frame silently collapses to a **0 ms window**.

I confirmed this against the real planner path (`window_exec_from_sql` → `get_window_length_ms`):

| Frame bound | Planned interval | `window_length_ms` |
|---|---|---|
| `INTERVAL '1000' MILLISECOND` | `{months: 0, days: 0, nanos: 1e9}` | `1000` |
| `INTERVAL '5' MINUTE` | `{months: 0, days: 0, nanos: 3e11}` | `300000` |
| `INTERVAL '7' DAY` | `{months: 0, days: 7, nanos: 0}` | `604800000` |
| `INTERVAL '1' MONTH` | `{months: 1, days: 0, nanos: 0}` | **`0`** |
| `INTERVAL '3' MONTH` | `{months: 3, days: 0, nanos: 0}` | **`0`** |
| `INTERVAL '1' YEAR` | `{months: 12, days: 0, nanos: 0}` | **`0`** |

The window length feeds retention (`retention_cutoff`), tiling eligibility (`TileConfig::usable_for_window_length_ms`, which requires `window_length_ms > 0`) and coverage geometry in `eval/advance.rs` / `eval/wro.rs`, so a month window silently produces wrong aggregates rather than failing loudly. This matters for the long-window use case in the README (`30d`, `1y`).

Note that the existing `_ => panic!("Unsupported window frame bound type")` arm shows unsupported bounds are meant to fail loudly, so returning `0` here is wrong either way.

## Fix

Add the `months` component, reusing `TimeGranularity` rather than repeating magic numbers:

```rust
(v.nanoseconds / 1_000_000)
    + (v.days as i64 * TimeGranularity::Days(1).to_millis())
    + (v.months as i64 * TimeGranularity::Months(1).to_millis())
```

The 30-day month matches the existing convention already used for tiles in `TimeGranularity::Months` (`model.rs`):

```rust
TimeGranularity::Months(m) => *m as i64 * 30 * 24 * 60 * 60 * 1000,
```

That choice is deliberate rather than cosmetic: tile boundaries are a fixed-size grid (`(timestamp / duration_millis) * duration_millis`), so using the same 30-day month keeps a month-based frame an exact multiple of `Months(1)` tiles. A calendar-accurate month (~30.44 days) would leave the frame misaligned with the tile grid.

Happy to switch to rejecting month intervals outright instead if you'd rather not imply calendar support — just say which you prefer.

### DIFF
--- a/src/runtime/operators/window/frame_utils.rs
+++ b/src/runtime/operators/window/frame_utils.rs
@@ -1,6 +1,10 @@
 use datafusion::logical_expr::{WindowFrame, WindowFrameBound, WindowFrameUnits};
 
+use crate::runtime::operators::window::model::TimeGranularity;
+
 /// RANGE window length in milliseconds.
+/// Month components use the same fixed 30-day month as [`TimeGranularity::Months`],
+/// so a month-based frame stays an exact multiple of month tiles.
 pub fn get_window_length_ms(window_frame: &WindowFrame) -> i64 {
     assert_eq!(
         window_frame.units,
@@ -10,7 +14,9 @@ pub fn get_window_length_ms(window_frame: &WindowFrame) -> i64 {
     match &window_frame.start_bound {
         WindowFrameBound::Preceding(value) => match value {
             datafusion::scalar::ScalarValue::IntervalMonthDayNano(Some(v)) => {
-                (v.nanoseconds / 1_000_000) + (v.days as i64 * 24 * 60 * 60 * 1000)
+                (v.nanoseconds / 1_000_000)
+                    + (v.days as i64 * TimeGranularity::Days(1).to_millis())
+                    + (v.months as i64 * TimeGranularity::Months(1).to_millis())
             }
             datafusion::scalar::ScalarValue::UInt64(Some(v)) => *v as i64,
             datafusion::scalar::ScalarValue::Int64(Some(v)) => *v,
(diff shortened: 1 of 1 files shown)


anovv:
Good catch, thanks! Can you make test suite smaller (1-2 tests) or omit it altogether, the logic doesn't seem to be too error prone

anovv:
I'm thinking if defaulting to 30-day month everywhere is a correct thing to do - get_window_length_ms may then return unexpected results (provided user may assume the real month, not the contract we silently agreed on in the code)? Should we not rely on TimeGranularity in get_window_length_ms at all and use whatever datafusion conversion logic is correct/default so we do not invent our own stuff (are there existing code examples/utilities in datafusion codebase of how to convert month intervals to ms properly that we can reuse?)? And also remove TimeGranularity::Month altogether to avoid messing up tiling, forcing user to always rely on days as a max granularity?

cheb0:
> And also remove TimeGranularity::Month altogether to avoid messing up tiling, forcing user to always rely on days as a max granularity

Look like a good choice to me.

anovv:
Ok, TimeGranularity::Month goes, INTERVAL MONTH stays, right? What's the conversion logic - 30 days?

Issue #247, opened 2026-08-20
Decouple streaming and request worker logic


anovv:
Currently Worker has all the logic for both streaming and request modes - we should have separate worker types

PR #285, opened 2026-09-06
Window store contract: operator namespace, key-group store clients per task

anovv:
## Summary
- Bind each WO task to an operator-scoped `StateNamespace` plus an owned `KeyGroupRange` (`max_p` is job-wide).
- Drop `namespace` from `stream_due` / `checkpoint` / `restore`; InMem restore and maintain isolate siblings by range, not a task-suffixed blob.
- Request-mode engine: HTTP processor only on the request-source worker, request source/sink colocation, WRO forwards request extras. Process-local `InMemory` is illegal in request mode.
- There is no InMemoryGrpc / fake remote store. Layer C lives on the last PR and talks to Scylla.

Stacked under: design #157. Next: `feat/scylla-wo-write` (#287).

## Test plan
- [ ] Two-task restore/maintain isolation test in `inmem.rs`
- [ ] Layer A + request-source unit tests


anovv:
## Review (correctness + simplify)

The contract is right: operator-scoped `StateNamespace`, range-bound clients, traits without `namespace` on `stream_due` / `checkpoint` / `restore`, and the two-task isolation test. A few copies of the same scope and a restore that filters instead of asserting same-assignment are what I’d cut.

### Correctness

**Restore can wipe the wrong task.** `restore_in` checks namespace, then keeps only partitions in *this* client’s range and deletes everything in that range. It never checks `checkpoint.key_group_range` / `max_parallelism` even though both are stored.

If task 1 is given task 0’s snapshot (same operator ns, different range): filter drops every partition, `retain(!owns)` clears task 1’s live data, and there is no error. Re-hashing keys with the *current* `scope.max_parallelism` has the same shape of bug if `max_p` changes.

v1 is same-assignment. Assert `range` and `max_p` match `scope`, then replace the owned range. Filter is a safety net, not the contract.

**`parallelism().max(1)` hides a bad task.** `range_for_subtask` already requires `p >= 1`. Coercing `0` to `1` makes that task own `range_for_subtask(0, 1, max_p)` — for `max_p > 1` that is the full key space — so maintain/restore would clobber siblings. Fail the same way as `task_index >= 0`.

**`Key::hash_from_bytes` is looser than `from_bytes`.** Short slices are zero-padded; `PartitionKey::key_group` then maps junk keys to group 0. Require 8 bytes:

```rust
u64::from_le_bytes(bytes[..8].try_into().expect("key bytes start with 8-byte hash"))
```

The other behavior changes look correct: WRO forwarding extras / ingest timestamp (request sink requires `SOURCE_REQUEST_ID` / `SOURCE_TASK_INDEX`), HTTP processor bound to local `vertex_ids`, and panicking on `OperatorPerWorker` + request I/O until #247.

The InMemory-in-request-mode check is a tripwire for later. Today the only `OperatorStateBackendConfig` is `InMemory` and `RequestStoreConfig` is empty, so request mode already cannot validate. Fine to keep; it does nothing until a shared backend exists.

### Simplify

`(namespace, max_p, range)` is stored three times: `WindowStoreTaskScope` on the client, `WindowOperatorState`, and `OperatorTaskState`. Maintenance already downcasts to `WindowOperatorState` for `retention_cutoff()`, then reads range off the generic trait.

Keep **one** `WindowStoreTaskScope` on `WindowOperatorState`. Drop `key_group_range` / `max_parallelism` from `OperatorTaskState`. In `maintain`, use the WO state’s scope. `OperatorStore::maintain`’s `ns` is the same value the registry already has on `state`.

Keep `writer_id` and `attempt` on the scope — that’s the single bind for every backend; InMem just doesn’t read them yet. When Scylla lands, fail closed if `execution_attempt_id` is missing rather than fencing with an empty token (`unwrap_or_default()` is fine for InMem).

That also shrinks `open()`: build one scope, pass it to `open_window_operator_store` and `WindowOperatorState::new`.

Smaller cuts:

- `maintain_cutoff` / `report_metrics`: take `&WindowStoreTaskScope` (or `owns_partition`) instead of `(ns, range, max_p)`
- `range_for_subtask` → `KeyGroupRange::for_subtask(...)`
- `InMemWindowStoreClient`’s `OperatorStore` impl is unused in production (registry holds the physical store)
- `RecordingWindowStore`: delegate all `WindowOperatorStore` methods through `client`, drop the extra `inner` handle
- `WindowOperatorState::for_test` → `new(..., WindowStoreTaskScope::for_test(ns))`

Keep the client wrapper (bind once, methods don’t take ns/range). Don’t copy that bind onto the task-state trait.

### Tests

`restore_and_maintain_do_not_clobber_sibling_task` is the test this PR needed. Two cheap extras:

1. Restoring snapshot 0 through client 1 errors (does not clear task 1).
2. `c0.load_key_state(&part1)` / `store_key_state` returns the out-of-range error.

anovv:
Addressed in d3cbff4 (stack rebased).

**Correctness** — all valid:
- Restore now asserts checkpoint `(namespace, key_group_range, max_parallelism)` matches the client, then replaces that range. Filter stays as a safety net. `restore_rejects_sibling_snapshot` covers snapshot 0 through client 1 (task 1 is not cleared).
- `open()` fails if `parallelism < 1` instead of `max(1)`.
- `Key::hash_from_bytes` requires 8 bytes (no zero-pad).
- `client_rejects_out_of_range_partition` covers `load_key_state` / `store_key_state` on a sibling key.

**Simplify** — done: one `WindowStoreTaskScope` on `WindowOperatorState`; `maintain_cutoff` / `report_metrics` take `&scope`; `KeyGroupRange::for_subtask`; `RecordingWindowStore` WO methods go through the bound client.

Left on purpose:
- `OperatorTaskState::key_group_range` / `max_parallelism` stay as thin delegates (Scylla maintain still reads them).
- Client `OperatorStore` impl and `RecordingWindowStore.inner` stay; inner is still the WRO / maintain handle.

InMemory-in-request-mode remains the tripwire for a shared backend.

anovv:
Follow-up on d3cbff4: the correctness fixes landed. Restore now requires same ns/range/`max_p` (sibling snapshot no longer wipes task 1), `open()` rejects `parallelism < 1`, `hash_from_bytes` requires 8 bytes, `WindowOperatorState` holds one `WindowStoreTaskScope`, maintain/metrics go through `owns_partition` / `wo.scope()`, and the extra isolation tests are in.

Leftover (optional, not blocking):

- `OperatorTaskState` still has `key_group_range()` / `max_parallelism()`. Nothing calls them — `maintain` uses `wo.scope()`. Drop them from the trait.
- `range_for_subtask` is now only a wrapper under `for_subtask`. Fold the body into the method and stop re-exporting the free function.
- `hash_from_bytes`: `bytes[..8]` panics on short slices before the `expect`. Use `bytes.get(..8)` if you want that message.
- `RecordingWindowStore.inner` is still needed for `WindowRequestStore` (unbounded reads). That split is correct.

I’d merge after dropping the unused trait methods.

anovv:
Follow-up addressed in fe19bb2.

- Dropped `key_group_range` / `max_parallelism` from `OperatorTaskState`. Maintain already had the window state's scope.
- Folded `range_for_subtask` into `KeyGroupRange::for_subtask` and stopped re-exporting the free function.
- `hash_from_bytes` uses `get(..8)` so the expect message fires on short slices.

`RecordingWindowStore.inner` stays for unbounded request reads.

Scylla maintain now reads the range from `wo.scope()` so the stack still compiles.

PR #287, opened 2026-09-06
[1/6] Scylla window store: session, write path, and load_triggers

anovv:
## Summary
- Depends on #296 (`load_triggers` + operator-owned due paging).
- Scylla session via `StateSessionHandle`: worker `configure` connects (keyspace DDL), `open_window_operator_store` takes the shared driver session.
- Unlogged versioned writes only: raw / tiles / `key_states` / triggers. Clustering includes `(attempt, E)`. No ingest LWT.
- **Dropped** `window_head` DDL, claim maps, and `recovery_bases`. Overlay / lease is #288.
- One CQL `AFTER` query per hop. Resume is a private `Seek` (last **raw** clustering + `(bucket, shard)`).

Next: `feat/scylla-wo-checkpoint` (#288). Prefetch later: #297.

Protocol: [#157](https://github.com/volga-project/volga/pull/157).

## Test plan
- [x] `next_partition_walks_time_buckets`
- [x] `trigger_partitions_are_stable_and_deduped`
- [ ] Ignored docker / testcontainer suite (`VOLGA_SCYLLA_CONTACT` or `scylladb/scylla:5.4`)

### DIFF
--- a/src/runtime/operators/window/store/backend/mod.rs
+++ b/src/runtime/operators/window/store/backend/mod.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 use anyhow::Result;
 use arrow::array::RecordBatch;
 use async_trait::async_trait;
-use futures::stream::BoxStream;
 use serde::{Deserialize, Serialize};
 
 use crate::api::spec::state::{OperatorStateBackendConfig, RequestStoreConfig};
@@ -16,8 +15,10 @@ use crate::runtime::state::{OperatorStore, StateRegistry};
 
 use super::WindowData;
 
+mod due;
 mod inmem;
 
+pub use due::{stream_due, trigger_fetch_limit, DueWorkStream};
 pub use inmem::{InMemWindowStore, InMemWindowStoreClient};
 
 /// Job-level execution attempt stamped on published versions.
@@ -98,7 +99,26 @@ pub struct DueWindowWork {
     pub triggers: Vec<WindowTrigger>,
 }
 
-pub type DueWorkStream<'a> = BoxStream<'a, Result<Vec<DueWindowWork>>>;
+/// Resume token for [`WindowOperatorStore::load_triggers`]. Opaque to the
+/// operator: only the backend that produced it should pass it back.
+#[derive(Debug, Clone)]
+pub struct TriggerResume {
+    pub last: WindowTrigger,
+    /// Last raw clustering `(attempt, epoch)` when the store distinguishes
+    /// overlay-hidden rows from visible ones. Empty attempt = last visible.
+    pub raw_attempt: Vec<u8>,
+    pub raw_epoch: i64,
+}
+
+impl TriggerResume {
+    pub fn after_visible(last: WindowTrigger) -> Self {
+        Self {
+            last,
+            raw_attempt: Vec::new(),
+            raw_epoch: 0,
+        }
+    }
+}
 
 /// Store operations used by the sole Window Operator for a partition.
 #[async_trait]
@@ -116,7 +136,13 @@ pub trait WindowOperatorStore: OperatorStore {
         meta: &KeyState,
         triggers: &[WindowTrigger],
     ) -> Result<()>;
-    fn stream_due<'a>(&'a self, after: Option<Cursor>, through: Cursor) -> DueWorkStream<'a>;
+    async fn load_triggers(
+        &self,
+        after: Option<Cursor>,
+        through: Cursor,
+        resume: Option<&TriggerResume>,
+        limit: usize,
+    ) -> Result<(Vec<WindowTrigger>, Option<TriggerResume>)>;
     async fn store_key_state(&self, partition: &PartitionKey, state: &KeyState) -> Result<()>;
     /// Complete all pending writes before capturing the returned snapshot.
     async fn checkpoint(&self) -> Result<WindowBackendSnapshot>;
--- a/config/runtime_consts.kube_test.json
+++ b/config/runtime_consts.kube_test.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.process_page_size": 256,
+  "window.process_page_size": 4096,
   "window.process_key_concurrency": 16
 }
--- a/config/runtime_consts.local_test.json
+++ b/config/runtime_consts.local_test.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "50ms",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.process_page_size": 256,
+  "window.process_page_size": 4096,
   "window.process_key_concurrency": 16
 }
--- a/config/runtime_consts.prod.json
+++ b/config/runtime_consts.prod.json
@@ -25,6 +25,6 @@
   "transport.tcp_connect_retry_delay": "1s",
   "window.ingest_max_records": 65536,
   "window.ingest_key_concurrency": 32,
-  "window.process_page_size": 256,
+  "window.process_page_size": 4096,
   "window.process_key_concurrency": 16
 }
--- a/src/runtime/consts.rs
+++ b/src/runtime/consts.rs
@@ -42,7 +42,7 @@ pub const TRANSPORT_TCP_CONNECT_RETRY_DELAY: &str = "transport.tcp_connect_retry
 pub const WINDOW_INGEST_MAX_RECORDS: &str = "window.ingest_max_records";
 /// In-flight keys for ingest `insert_batch` and WRO `process_key`.
 pub const WINDOW_INGEST_KEY_CONCURRENCY: &str = "window.ingest_key_concurrency";
-/// Triggers per `stream_due` page (eval + emit before the next page).
+/// Trigger fetch cap for one due hop (`max` with process key concurrency).
 pub const WINDOW_PROCESS_PAGE_SIZE: &str = "window.process_page_size";
 /// In-flight keys evaluating one process page.
 pub const WINDOW_PROCESS_KEY_CONCURRENCY: &str = "window.process_key_concurrency";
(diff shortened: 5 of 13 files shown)


anovv:
Follow-ups for later PRs (not this one):

- Head LWT is #288 — this PR still last-write-wins upserts `window_head`.
- Foyer so the hot path can `IF writer_epoch = prev` from a cached writer_version.
- Split writer vs serving until recovery catch-up (WRO pin).
- `USING TIMESTAMP` only if LWT p99 still dominates after Foyer + one-shot LWT.

src/runtime/operators/window/store/backend/mod.rs
```diff
@@ -116,6 +141,7 @@ impl TriggerResume {
             last,
             raw_attempt: Vec::new(),
             raw_epoch: 0,
+            part_idx: 0,
         }
     }
```
anovv, line 156: This is the leak from #296, now worse: `part_idx` with “InMem ignores this.” The helper must not interpret the token. Private newtype; this file should not name overlay clustering or shard indexes.

Scylla resume is last **raw** row + `(bucket, shard)`, not last visible `WindowTrigger`. `after_visible` + empty `raw_attempt` is a second mode (`seek_from_resume` returns `None`) that exists only because the type is shared.

src/runtime/operators/window/store/backend/scylla/triggers.rs
```diff
@@ -0,0 +1,317 @@
...
+    seek: Option<&TriggerSeek>,
+) -> Option<TriggerResume> {
+    if part_idx >= part_count {
+        return None;
+    }
+    match seek {
+        Some(seek) => Some(resume_from_seek(ns, part_idx, seek)),
+        None => Some(resume_at_shard(ns, after, part_idx)),
+    }
+}
```
anovv: `seek_from_resume` / `resume_from_seek` / `resume_at_shard` / `resume_token` exist to pack a raw clustering key into a `WindowTrigger` and to treat empty attempt as “start of shard.” That is most of this file.

Store a `Seek { bucket, shard, fire_ts, fire_seq, business_key, kind, window_id, attempt, epoch }` and always use `SELECT_TRIGGERS_AFTER`. New shard = next `(bucket, shard)` + min seek from `after`. Then these four functions go away.

anovv (commented):
Right split: no store-owned `stream_due`, one `load_triggers` hop from the #296 helper. The flatten (LIMIT = remaining visible so a hidden page cannot skip later rows) is the correct *store* responsibility if the helper treats empty as EOS.

That is also the extra complexity. You are paging twice — helper until empty, Scylla until `selected.len() == limit` — then encoding the position back through a shared `TriggerResume` that looks like a `WindowTrigger`.

Simplest system, less code:

1. **One CQL query per `load_triggers`.** Always `SELECT ... AND clustering > ? AND fire_ts <= through LIMIT ?`. First hop / new shard: seek is `after` (or zeros). Mid-shard: last **raw** row. Drop `SELECT_TRIGGERS`, `resume_at_shard`, and empty-`raw_attempt` meaning “no seek.”
2. **Do not fill `limit` visible inside the store.** Return this page’s visible rows (maybe empty) + `next`. Helper continues while `next` is `Some`. EOF is `next is None` (needs a small #296 contract change).
3. **Token is Scylla-private** (`bucket`, `shard`, raw clustering). Not `part_idx` into a rebuilt `Vec`, not `WindowTrigger.last`. `next_partition(bucket, shard)` instead of materializing the whole `(after, through]` grid.

Keep: overlay filter, raw seek including attempt/epoch, `(bucket, kg_shard)` walk. Those are the real backend bits.

`SELECT_TRIGGERS` uses `fire_ts >= after.ts` while visibility is `fire_at > after`. With the operator’s `after = Cursor(frontier, u64::MAX)` same-ts rows are fetched then dropped. Always-AFTER from the cursor removes that.

src/runtime/operators/window/store/backend/scylla/triggers.rs
```diff
@@ -0,0 +1,317 @@
...
+        if let Some(next) = last_raw {
+            seek = Some(next);
+        }
+        if raw_len < remaining as usize {
+            part_idx += 1;
+            seek = None;
+        } else if raw_len == 0 {
+            break;
+        }
+    }
```
anovv, line 226: Second pager: loop until `limit` visible, CQL `LIMIT = remaining`. Needed only because #296 stops on empty triggers.

One query, return `(visible, next)` even if `visible` is empty. Helper keeps going while `next` is `Some`. Deletes `remaining`, the inner `while`, and the dead `else if raw_len == 0` (that branch cannot fire: `remaining >= 1` from the `while` condition).

`part_idx` is an index into a `Vec` rebuilt from `after`/`through` every call. Safe only because the helper never changes the range. Token should hold `(bucket, shard)` and `next_partition` should advance; no full grid.

src/runtime/operators/window/store/backend/scylla/cql.rs
```diff
@@ -0,0 +1,55 @@
...
+pub(super) const SELECT_RAW: &str = "SELECT event_ts, seq_no, attempt, epoch, payload FROM window_raw WHERE namespace = ? AND key_group = ? AND business_key = ? AND bucket_start = ? AND event_ts >= ? AND event_ts <= ?";
+pub(super) const SELECT_TILES: &str = "SELECT tile_start, attempt, epoch, payload FROM window_tiles WHERE namespace = ? AND key_group = ? AND business_key = ? AND granularity_ms = ? AND bucket_start = ? AND tile_start >= ? AND tile_start < ?";
+pub(super) const SELECT_TRIGGERS: &str = "SELECT fire_ts, fire_seq, business_key, trigger_kind, window_id, key_group, attempt, epoch FROM window_triggers WHERE namespace = ? AND bucket_start = ? AND kg_shard = ? AND fire_ts >= ? AND fire_ts <= ? LIMIT ?";
+pub(super) const SELECT_TRIGGERS_AFTER: &str = "SELECT fire_ts, fire_seq, business_key, trigger_kind, window_id, key_group, attempt, epoch FROM window_triggers WHERE namespace = ? AND bucket_start = ? AND kg_shard = ? AND (fire_ts, fire_seq, business_key, trigger_kind, window_id, attempt, epoch) > (?, ?, ?, ?, ?, ?, ?) AND fire_ts <= ? LIMIT ?";
```
anovv: Two statements for “first page of shard” vs “after seek.” One `SELECT ... (fire_ts, fire_seq, business_key, trigger_kind, window_id, attempt, epoch) > ? AND fire_ts <= ? LIMIT ?` covers both if the first seek is the watermark cursor.

`fire_ts >= ?` does not match `fire_at > after` (cursor). Same-ts rows are pulled and discarded in `visible_trigger`.

src/runtime/operators/window/store/backend/mod.rs
anovv: Done. Trait token is opaque. Scylla resume is a private `Seek` via `TriggerResume::opaque`; this file does not pack a `WindowTrigger`.

src/runtime/operators/window/store/backend/scylla/triggers.rs
anovv: Done. One `SELECT ... AFTER` query. `Seek` holds last raw clustering + `(bucket, shard)`. New shard is `next_partition` + `min_seek(after)`.

src/runtime/operators/window/store/backend/scylla/triggers.rs
anovv: Done. One query per hop. Empty visible + `Some(resume)` is legal; the operator continues while `next` is `Some`. No fill-`limit` loop, no `part_idx` grid.

src/runtime/operators/window/store/backend/scylla/cql.rs
anovv: Done. Dropped `fire_ts >=`. First hop / new shard seek from the watermark cursor so same-ts rows are not pulled.

anovv:
**Protocol change on #157 — keep the write path, drop head/claim work.**

Normative: [#157 *Protocol (normative)*](https://github.com/volga-project/volga/pull/157) and [`docs/window-store-impl.md`](https://github.com/volga-project/volga/blob/docs/window-store-design/docs/window-store-impl.md).

**Keep:** session, DDL for versioned raw / tiles / `key_states` / **triggers**, unlogged `commit_events`, `load_triggers` (one AFTER query + last-raw seek), derive `key_group`. Overlay is `me` until checkpoint exists, then `me | (cp.attempt ∧ E ≤ cp_E)`.

**Drop / do not add here:**
- `window_head` DDL (replaced by request-only `window_kg_lease` in #288)
- HeadClaim, DashMaps, per-key caches, ingest LWT
- `window_recovery_bases`

`E` is monotonic per `key_group` (or one shared counter for owned groups) and **must not restart at 0**. Serving steal/promote is not this PR.

anovv:
Landed on the latest #157 protocol.

Kept: session, versioned unlogged writes (`attempt`, `E` in clustering), `load_triggers` (one AFTER query + last-raw seek). Dropped `window_head` DDL, claim maps, and `recovery_bases`. Overlay / `window_kg_lease` is #288.

anovv:
Overlay changed again on #157: **`me | E ≤ cp_E` (any attempt)**, not `me | (cp.attempt ∧ E ≤ cp_E)`.

A single checkpoint blob is a cutoff, not a generation. The old predicate dropped idle keys after a second failover. Until a checkpoint exists, `me` only is still fine.

anovv:
Ignore the previous overlay comment. **#287 is `me` only** — no checkpoint restore yet.

Overlay `me | (cp.attempt ∧ E ≤ cp_E)` lands in **#288**, not here. Do **not** implement `me | E ≤ cp_E` (any attempt).

anovv:
Addressed on `da56fb09` (restacked through #293).

**Fixed**
- `load_tiles` / `load_raw` walk every 60s bucket via `time_buckets`. Last included ts is `to.ts` when `seq_no > 0`, else `to.ts - 1`. Docker roundtrip covers `[Cursor(50_000, 0), Cursor(60_000, 2))` and a tile run `[1000, 120000)`.
- Due walk is clamped: first tick starts at `through`'s bucket; close (`through == i64::MAX`) stays on the frontier bucket. No paging.
- `datacenter` uses `SessionBuilder::prefer_datacenter`.
- `process_due_emits_pages_before_watermark` expects one page (one key = one `advance_key`).

**Simplified**
- Deleted unused `encode_batches` / `decode_batches`.
- Inlined `collect_triggers` as `load_triggers`; dropped `due.rs`.

**Left as specified**
- Failed `commit_events` still fails the task (no caller retry, no new epoch).
- Kept `cut_top`, in-flight epochs, me-only overlay, full `key_state` read, `(fire_ts, fire_seq)` seek, `window_triggers` name.

Follow-up for a long but finite catch-up returning too many rows: #309.

PR #288, opened 2026-09-06
[2/6] Scylla window store: versioned checkpoint and restore

anovv:
## Summary
- Restack onto [#157](https://github.com/volga-project/volga/pull/157): publication **is** the completed checkpoint. No lease, steal, `serving_publish`, or `window_recovery_bases`.
- `WindowBackendSnapshot::Versioned { attempt, range, cuts }` — cuts parallel to the bound key-group range. Epoch restarts at 0 after restore; `cp_cut` is the restored overlay.
- `window_kg_meta` (request mode only), **two** statements:
  - **(A) publish** — `prev_cut = row.cut`, `IF cur_attempt <= me AND checkpoint_id = observed`. First row is `INSERT IF NOT EXISTS` from a publish, never from a fresh-job open.
  - **(B) take_attempt** — `SET cur_attempt = me IF cur_attempt <= me`. Not (A) with an unchanged cut.
- Trigger 1: `notify_checkpoint_complete` (operator hook; master delivery is #300). Trigger 2: heal at open if `row.checkpoint_id < restored_id`, else (B). Must finish before ingest.

Depends on #287. Do not merge until the stack is complete.

## Test plan
- [ ] InMem still restores same-assignment snapshots; rejects `Versioned`
- [ ] Ignored docker: `scylla_restore_sees_checkpointed_prefix`
- [ ] Ignored docker: overlay still hides an unrestored attempt (`scylla_overlay_hides_other_attempt`)

PR #289, opened 2026-09-06
[3/6] Scylla window store: maintain / GC

anovv:
## Summary
- Restack onto [#288](https://github.com/volga-project/volga/pull/288) / [#157](https://github.com/volga-project/volga/pull/157). No lease / `prev_E` extras.
- Physical delete is gated on the **committed** watermark (last completed checkpoint), never the live one. Streaming seeds it at restore and advances it on `notify_checkpoint_complete`.
- Per-cell version retention, three slots: `my_attempt` (streaming) or `window_kg_meta.cur_attempt` (request), plus `cut` and `prev_cut`. Zombies above a cut entry are collected.
- Expired `window_kg_buckets` scans drop matching raw/tile partitions. Consumed-due GC waits on the deferred due-bucket index.

Depends on #288. Do not merge until the stack is complete.

## Test plan
- [ ] InMem maintain no-ops until a checkpoint completes; retention tests complete a checkpoint first
- [ ] `keep_slots` unit test: mine / cut / prev, drop zombies
- [ ] Ignored docker: bucket delete + version GC (follow-up on this PR if not already)


anovv:
Follow-ups for later PRs:

- Batch the DELETEs instead of one request per bucket/granularity.
- TWCS/TTL instead of per-bucket DELETE where that matches retention.
- Foyer for the `window_kg_buckets` index.

anovv:
**Wait on the #288 rewrite; GC rules changed with the protocol on #157.**

Keep event-time bucket GC (`window_kg_buckets` / `floor_bucket`, raw + all tile granularities).

**Also:** drop displaced extras `attempt == prev.attempt AND E > prev.E` (from the group lease, not a bases table).

**Do not:**
- GC from `window_recovery_bases` / per-key `window_head`
- drop unpublished current-writer rows (`E > serving_E`)
- drop `E ≤ serving_E` holes until time GC / compact
- skip serving-pin grace for in-flight WRO

See *State prune / cleanup* in [#157](https://github.com/volga-project/volga/pull/157).

anovv:
GC follows the new protocol: event-time buckets as before, plus drop `prev.attempt ∧ E > prev.E`. Unpublished current-writer rows and `E ≤ serving_E` holes stay until time GC / compact.

anovv:
GC drop is **`E > prev_E AND attempt != owner`**, not `prev.attempt ∧ E > prev.E`.

Keep unpublished current-writer rows and `E ≤ prev_E` holes until event-time GC. Grace is for **time GC vs in-flight WRO windows below `data_floor`**, not a pin refcount — extras above `prev_E` are already filtered on the read path.

src/runtime/operators/window/state.rs
```diff
@@ -203,15 +244,18 @@ impl WindowOperatorState {
...
+        };
 
         let mut key_state = self
             .store
             .load_key_state(&partition)
             .await
             .expect("key state");
-        let (accepted, dropped) =
-            drop_late_entries(&batch, self.ts_column_index, watermark_frontier);
+        let (accepted, dropped) = drop_late_entries(&batch, self.ts_column_index, cutoff);
```
anovv, line 273: do we need to drop lates for request mode (StateOnly)?

PR #290, opened 2026-09-06
[4/6] Scylla request store: ReadOptions, given session, meta pin

anovv:
## Summary
- `load_window_data` takes `ReadOptions { Committed, Fresh }` and returns `WindowRead` (`data`, `committed_wm`, `checkpoint_id`, `retention_floor`).
- The store is **given** a session and `max_parallelism`; it does not call `connect()`, and it has no key-group range or lifecycle methods.
- Each lookup pins `window_kg_meta` (not cached in v1). Absent/empty cut returns empty. Coverage refuses `lo < retention_floor`. After the data read, re-read meta and fail on `pinned_id < row.prev_checkpoint_id`.
- `Fresh` is head-scoped: `[lo, committed_wm]` at the cut; `(committed_wm, hi]` raw with `cut.allows(v) || v.attempt == cur_attempt`.

Depends on #289. Next: #292.

`#156` and `#300` still block request-mode Scylla **open**. Protocol: [#157](https://github.com/volga-project/volga/pull/157). Do not merge this stack.

## Test plan
- [ ] InMem WRO path still answers through `ReadOptions::Committed` (empty meta fields)
- [ ] Coverage guard refuses below `retention_floor`
- [ ] Scylla `Committed` sees only the published cut; `Fresh` sees the current-attempt head
- [ ] Staleness re-read fails when the pin aged out of `prev_checkpoint_id`


anovv:
Follow-ups for later PRs:

- Keep serving behind writer until recovery catch-up so WRO does not observe an incomplete writer (today `publish_head` sets both together).
- Foyer for the head row / serving pin.

anovv:
**WRO pin is the group lease, not `window_head.serving_*` as a per-key head, and not writer overlay.**

After #288 rewrite: point-get `window_kg_lease` **once** per request, then every cell:

```text
E <= serving_E
AND NOT (attempt == prev.attempt AND E > prev.E)
```

Then latest per cell: same `Cursor` / `(granularity, tile_start)` / `key_state` → `max(E)`, then `max(attempt)`. Filter is **client-side** (time-first clustering).

Do **not**:
- mix writer overlay with serving
- use epoch-only `E ≤ serving_E` (zombie leak after steal)
- use `attempt == serving.attempt` only (holes on older attempts disappear)
- read triggers on WRO
- reread the lease mid-request

Layer C oracle = last **serving** publish, with grey window until steal accepted. Details: [#157](https://github.com/volga-project/volga/pull/157) *WRO reads*.

anovv:
WRO pins `window_kg_lease` once, then filters client-side:

```
E ≤ serving_E AND NOT (attempt == prev.attempt AND E > prev.E)
```

then `max(E)` / `max(attempt)` per cell. Same data tables as streaming. Layer C runner restored on Scylla.

anovv:
WRO filter on #157 is now:

```text
E <= serving_E
AND (prev_E unset OR E <= prev_E OR attempt == serving.attempt)
```

Pin `(serving_*, prev_E)` once. Do **not** use `NOT (attempt == prev.attempt AND E > prev.E)` — that leaks B if C steals before B publishes.

Then per-cell `max(E)`, then `max(attempt)`.

anovv:
WRO filter is unchanged (the right one):

```text
E <= serving_E
AND (E <= prev_E OR attempt == serving.attempt)
```

`prev_E` is 0 in CQL. Empty serving ⇒ **no pin**, return empty. Do not treat `prev_E = 0` as a snapshot.

src/runtime/operators/window/store/backend/scylla/request.rs
```diff
@@ -0,0 +1,277 @@
...
+        kg: i32,
+        raw_runs: &[RawRun],
+        tile_runs: &[TileRun],
+        is_visible: impl Fn(Version) -> bool + Send + Sync,
+    ) -> Result<(Vec<arrow::array::RecordBatch>, TileMap)> {
+        let prepared = self.prepared().await?;
+        let session = Arc::clone(&self.session);
+        let ns = partition.namespace.clone();
+        let key = partition.business_key.clone();
+        let raw = fetch_raw(
```
anovv, line 104: parallelize fetch_raw and fetch_tiles?

PR #296, opened 2026-09-12
Move due paging into the operator via load_triggers

anovv:
## Summary
- `load_triggers` is one hop. The operator loops it in `emit_due_pages`.
- `TriggerResume` is a private newtype. The helper does not read it. InMem stores the last visible `WindowTrigger`.
- EOF is `next is None`. Empty `triggers` + `Some(resume)` is a legal hop. The store does not fill `limit` visible rows.
- Load+eval share the `process_key_concurrency` (16) pool. `process_page_size` stays 256 until Scylla is measured.
- Same-`fire_at` resume is full `WindowTrigger` order (`trigger > *resume.last()`).

Prefetch later: #297.

## Test plan
- [x] InMem store unit tests, including same-`fire_at` across keys
- [x] Window tests including multi-page due emit

### DIFF
--- a/src/runtime/operators/window/store/backend/due.rs
+++ b/src/runtime/operators/window/store/backend/due.rs
@@ -1,0 +1,68 @@
+use std::collections::BTreeMap;
+
+use anyhow::Result;
+use futures::future::try_join_all;
+use futures::stream::BoxStream;
+
+use crate::runtime::consts::{
+    runtime_consts, WINDOW_PROCESS_KEY_CONCURRENCY, WINDOW_PROCESS_PAGE_SIZE,
+};
+use crate::runtime::operators::window::model::{Cursor, PartitionKey, WindowTrigger};
+
+use super::{DueWindowWork, TriggerResume, WindowOperatorStore};
+
+pub type DueWorkStream<'a> = BoxStream<'a, Result<Vec<DueWindowWork>>>;
+
+/// Visible triggers to fetch per store hop: page cap, at least the key pool.
+pub fn trigger_fetch_limit() -> usize {
+    let page = runtime_consts().u64(WINDOW_PROCESS_PAGE_SIZE).max(1) as usize;
+    let concurrency = runtime_consts().u64(WINDOW_PROCESS_KEY_CONCURRENCY).max(1) as usize;
+    page.max(concurrency)
+}
+
+pub async fn group_due_work(
+    store: &dyn WindowOperatorStore,
+    selected: Vec<WindowTrigger>,
+) -> Result<Vec<DueWindowWork>> {
+    let mut grouped: BTreeMap<PartitionKey, Vec<WindowTrigger>> = BTreeMap::new();
+    for trigger in selected {
+        grouped
+            .entry(trigger.partition.clone())
+            .or_default()
+            .push(trigger);
+    }
+    try_join_all(grouped.into_iter().map(|(partition, triggers)| async move {
+        let key_state = store.load_key_state(&partition).await?;
+        Ok(DueWindowWork {
+            partition,
+            key_state,
+            triggers,
+        })
+    }))
+    .await
+}
+
+/// Operator-owned due stream: store only pages triggers.
+pub fn stream_due<'a>(
+    store: &'a dyn WindowOperatorStore,
+    after: Option<Cursor>,
+    through: Cursor,
+) -> DueWorkStream<'a> {
+    let limit = trigger_fetch_limit();
+    Box::pin(futures::stream::try_unfold(
+        Some(None::<TriggerResume>),
+        move |state| async move {
+            let Some(resume) = state else {
+                return Ok(None);
+            };
+            let (triggers, next) = store
+                .load_triggers(after, through, resume.as_ref(), limit)
+                .await?;
+            if triggers.is_empty() {
+                return Ok(None);
+            }
+            let work = group_due_work(store, triggers).await?;
+            Ok(Some((work, next.map(Some))))
+        },
+    ))
+}
--- a/src/runtime/operators/window/store/backend/inmem.rs
+++ b/src/runtime/operators/window/store/backend/inmem.rs
@@ -16,7 +16,6 @@ use serde::{Deserialize, Serialize};
 
 use std::any::Any;
 
-use crate::runtime::consts::{runtime_consts, WINDOW_PROCESS_PAGE_SIZE};
 use crate::runtime::metrics::MetricsLabels;
 use crate::runtime::operators::window::metrics;
 use crate::common::KeyGroupRange;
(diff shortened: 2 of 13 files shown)


src/runtime/operators/window/store/backend/mod.rs
```diff
@@ -98,7 +99,26 @@ pub struct DueWindowWork {
...
+}
+
+impl TriggerResume {
+    pub fn after_visible(last: WindowTrigger) -> Self {
+        Self {
+            last,
+            raw_attempt: Vec::new(),
+            raw_epoch: 0,
+        }
+    }
```
anovv, line 115: `Opaque to the operator` but the fields are public and already Scylla-shaped. InMem only uses `last`. #287 adds `part_idx` and treats empty `raw_attempt` as “no seek,” so this type is already incomplete and leaky.

Private newtype + crate constructors. InMem stores last visible `WindowTrigger`. Scylla stores last **raw** clustering + shard. Do not freeze overlay fields on the trait in this PR — last visible as the token is wrong under overlay, but stuffing Scylla seek into a `WindowTrigger` is worse.

src/runtime/operators/window/store/backend/due.rs
```diff
@@ -0,0 +1,68 @@
...
+            let (triggers, next) = store
+                .load_triggers(after, through, resume.as_ref(), limit)
+                .await?;
+            if triggers.is_empty() {
+                return Ok(None);
+            }
+            let work = group_due_work(store, triggers).await?;
+            Ok(Some((work, next.map(Some))))
+        },
+    ))
```
anovv: This `try_unfold` + `Some(None::<TriggerResume>)` is leftover store-stream shape. Paging is now the operator’s job — a loop in `emit_due_pages` is the same API with less machinery.

Also: empty `triggers` stops even if `next` is `Some`. That only works if every backend fills visible rows internally. Prefer **EOF = `next is None`** so a hidden-only hop can continue. Short pages are fine; do not make Scylla look like InMem’s full `take(limit)`.

src/runtime/operators/window/store/backend/due.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+    try_join_all(grouped.into_iter().map(|(partition, triggers)| async move {
+        let key_state = store.load_key_state(&partition).await?;
+        Ok(DueWindowWork {
+            partition,
+            key_state,
+            triggers,
+        })
+    }))
+    .await
```
anovv: `try_join_all` loads every key in the page with no cap. `process_key_concurrency` only applies later in `emit_due_pages`. With `process_page_size` 4096 this is 4096 `load_key_state`s before eval — fine for InMem, not for #288 owner-steal LWT.

Group in the operator and load+eval on the same 16-wide pool. Then this helper can go away.

`trigger_fetch_limit` as `page.max(concurrency)` is dead at 4096 vs 16. Use `process_page_size` and leave it at 256 until you measure.

src/runtime/operators/window/store/backend/inmem.rs
```diff
@@ -381,60 +380,34 @@ impl InMemWindowStore {
...
+                resume
+                    .map(|resume| trigger > &&resume.last)
+                    .unwrap_or(true)
+            })
+            .take(limit)
+            .cloned()
+            .collect::<Vec<_>>();
+        let next = (selected.len() == limit)
+            .then(|| selected.last().cloned().map(TriggerResume::after_visible))
+            .flatten();
```
anovv, line 409: This is the simple hop: filter, `take(limit)`, resume = last visible. Keep it. The comparison is `trigger > &&resume.last` — `**trigger > resume.last` is enough.

Worth a unit test: two keys, same `fire_at`, `limit = 1`. Resume must be full `WindowTrigger` order, not cursor-only, or the second key is skipped.

src/runtime/operators/window/tests/smoke.rs
```diff
@@ -199,21 +200,23 @@ WINDOW w AS (
   ORDER BY timestamp
   RANGE BETWEEN INTERVAL '5000' MILLISECOND PRECEDING AND CURRENT ROW
 )"#;
-    let n = 300i64;
+    let n = (trigger_fetch_limit() + 34) as i64;
```
anovv: InMem `due_triggers_are_range_filtered_and_backend_paged` already proves multi-page. Tying smoke size to `trigger_fetch_limit() + 34` (4130 rows) makes a correctness smoke a page-size test. Keep `n` small.

anovv (commented):
The cut is right: helper pages, store only `load_triggers`. #287 is already using it that way.

What I would change before this becomes the Scylla API:

1. **One pager.** `stream_due` is the only loop. `load_triggers` is one hop (InMem already is). Do not require the store to fill `limit` visible rows — short pages and empty+`Some(resume)` must be legal, and **EOF is `next is None`**. Write that on the trait. Today the helper treats empty as EOS, which is only safe if every backend loops internally (Scylla does, and that is the extra code).
2. **`TriggerResume` is not opaque.** Public `raw_attempt` / `raw_epoch` are Scylla overlay fields; #287 then adds `part_idx`. InMem only needs last visible trigger. Make this a private newtype. The helper must not read it.
3. **Drop the stream wrapper and the unbounded key-state join.** A loop in `emit_due_pages` + `buffered(process_key_concurrency)` for load+eval is less code and avoids 4096 `load_key_state`s before eval. Keep `process_page_size` at 256 until Scylla is measured; `max(concurrency)` never fires.

Prefetch (#297) stays out until this is a boring loop.

src/runtime/operators/window/store/backend/mod.rs
anovv: Done. `TriggerResume` is a private newtype (`after_visible` / `opaque`). Helper does not read it.

src/runtime/operators/window/store/backend/due.rs
anovv: Done. `emit_due_pages` loops `load_triggers`. EOF is `next is None`; empty + `Some(resume)` continues.

src/runtime/operators/window/store/backend/due.rs
anovv: Done. Operator groups, then `buffered(concurrency)` `advance_key` (loads `key_state` itself). Page size back to 256.

src/runtime/operators/window/store/backend/inmem.rs
anovv: Kept the simple hop (`**trigger > *resume.last()`). Added `load_triggers_resumes_same_fire_at_across_keys`. A full last hop may still return a resume; the next empty hop is EOF.

src/runtime/operators/window/tests/smoke.rs
anovv: Done. Smoke `n` is 300 again.

PR #298, opened 2026-09-18
Document revised Scylla window store protocol

anovv:
## Summary
- New design PR for the Scylla window store. **Supersedes the protocol in #157** (`window_kg_lease` / `prev_E` / pin-from-serving-E).
- **Streaming:** unlogged versioned writes; WO overlay `me | (cp_g.attempt ∧ E ≤ cp_E_g)`. Checkpoint blob is per-owned-group `cp_E` so rescale is slice remap, not a data copy. No pins, no ingest LWT.
- **Request:** `window_pins` (static `owner` / `serving_wm`, per-key `serving_attempt` / `serving_E`). Steal owner only; idle keys keep their pin. WRO: `attempt == serving_attempt ∧ E ≤ serving_E`.
- **Publish gate:** `current_wm >= serving_wm ∧ current_wm > cp_wm ∧ dirty ≠ ∅`. Not on the CP barrier. Publish CAS miss fails the task (backup to master kill).

Normative cut: *Protocol (normative)* in [`STORE_DESIGN.md`](https://github.com/volga-project/volga/blob/docs/window-store-protocol/src/runtime/operators/window/STORE_DESIGN.md). Stack notes: [`docs/window-store-impl.md`](https://github.com/volga-project/volga/blob/docs/window-store-protocol/docs/window-store-impl.md).

## Test plan
- [x] Documentation-only change
- [ ] Restack #287–#290 onto this protocol after merge (`window_pins`, per-group CP, `publish_ok`; drop `prev_E`)


anovv:
## Review

The overlay revert and per-key pins are the right direction vs #157. Do not restack #287–#290 onto this until the republish history cut is closed.

Idle keys are solved. **Touched keys are not.**

---

### Blocking: first publish of a key drops A’s history

WRO is:

```text
attempt == serving_attempt AND E ≤ serving_E
```

Idle keys keep their clustering pin — good. The moment a key is in `dirty` and `publish_ok` fires, `serving_attempt` becomes `B` and **every A row for that key disappears**. The doc already says this:

> as soon as `serving_attempt` becomes `B`, that key’s A rows disappear
> First B publish may be `B@150` replacing `A@5000` on keys in that batch

Source restore replays **post-CP** input only. Pre-CP raw/tiles stay as `A` under overlay. B’s `commit_events` writes new events + affected tiles, not the retained window. `> cp_wm` is a 1ms-past-restore watermark gate, not “catch-up complete.”

So the first OnCommit after replay emits one watermark can publish a prefix: WRO rebuilds a 1-hour window from B’s new rows only. That is a wrong answer, not a stale one.

`prev_E` existed for this: mix `A@80` and `B@160` **cells**, never two payloads for the **same** cell. Per-key pins remove the need for a **group-wide** `prev_E`. They do not remove hole-fill for a **republished** key.

Pick one, write it as normative:

1. **Per-key prev** (cheap, matches the new table): on publish, `prev_* ← serving_*`, then set serving to me. WRO:

   ```text
   (attempt == serving.attempt AND E ≤ serving_E)
   OR (attempt == prev.attempt AND E ≤ prev_E)
   ```

   Idle keys never get a `prev`. Touched keys keep A’s cut as holes until time GC.

2. **Copy-on-first-touch:** before a key may enter `dirty`, rewrite every retained cell under `me`. Then `attempt == serving.attempt` is complete. That is copy-on-restore, scoped to dirty keys — say so, including cost.

3. **Do not publish a key until B’s rows cover the window.** Same cost as (2), easy to get wrong.

“Not `prev_E` hole-fill because idle keys keep their pin” only argues (idle). It does not argue (republish).

GC currently drops anything not in overlay ∪ serving pin. After a prefix publish, that **deletes** the A history WRO just lost. If you add per-key `prev`, GC must keep `prev` rows too.

---

### What landed correctly

- WO overlay is `me | (cp_g.attempt ∧ E ≤ cp_E_g)` again. Do **not** go back to any-attempt `E ≤ cp_E`.
- Per-group `cp_E[]` so a slow group’s post-CP writes are not swallowed by `max_E`.
- `E` seeded from `cp_E_g`, not from `serving_E`. Numeric `B@150` vs `A@5000` is fine **across attempts**; WRO must not `max(E)` across attempts.
- Steal does not move clustering pins. Grey window until steal is accepted.
- Publish is not on the CP barrier. `>= serving_wm` **and** `> cp_wm` is the right pair for the three restore cases in the table.
- Streaming still has no owner / ingest LWT.

Documented WO cost is still true: after B checkpoints, C’s overlay is `B`, so idle keys last written by A disappear from **WO**. WRO keeps them via pins. That split is OK if you accept it.

You already pay `O(groups)` in the blob. Storing per-group `(attempt, E)` (inherit A’s cut for untouched groups) would keep those keys on WO too, without `recovery_bases`. Optional; not a substitute for the WRO hole.

---

### Spec gaps (fix in this PR)

**Checkpoint of untouched groups.** Blob is `attempt = this writer` plus `cp_E[g]` for every owned group. If B never wrote `g`, `next_E[g]` is unset. Persist the **inherited** slice cut, not `0`. Overlay still will not see A after C restores (attempt is B); this only keeps B’s own overlay correct (`B | A≤inherited` while B is live).

**`window_pins` is the hot partition you forbade on data tables.** `PRIMARY KEY ((namespace, key_group), business_key)` is one partition for every key in the group. Data tables refused that because a group is unbounded. Pins need a cardinality story: shard, or per-key partitions + a separate owner row.

**One Paxos BATCH cannot flush unbounded `dirty`.** Scylla LWT batches have size limits. Chunk publishes. WRO only reads per-key `serving_*`, so chunking is safe; `serving_wm` can move separately. Do not spec “statics + all dirty keys, one Paxos.”

**Steal fanout.** Restore does `SET owner` per owned group before ingest. `O(max_p / p)` LWTs, fail the task on miss/timeout. Say they run in parallel. Read-then-CAS (`IF owner = previous`); do not invent previous from the CP blob.

**Bootstrap.** First start (no restore): `cp_wm` unset, `serving_wm` unset, overlay `me` only. Spell out `publish_ok` (skip both wm compares; require `current_wm` set and `dirty ≠ ∅`).

**Pin GC.** Event-time data GC does not drop clustering pin rows. Expired keys leave pins that WRO point-gets into empty data. Drop pin rows for keys whose data is gone.

**`store_key_state`.** Does it mark `dirty`? WRO does not read `key_state` today; say so.

---

### Nits

- `> cp_wm` is not a catch-up fence. If replayed events generate `wm > cp_wm` on the first batch, `publish_ok` opens immediately. That is the blocking issue above, not a table error.
- Watermarks are non-decreasing: if the frontier stalls at `cp_wm`, pins never move. Conservative and fine.
- `VersionedRange` markdown is truncated (`Vec `). Fix the type.
- Failover ASCII: `WO overlay A≤100 ∪ B` is right; WRO `A@5000 until B publish_ok` is right only for **idle** keys.

---

### Restack

After the republish rule is in *Protocol (normative)*: drop `window_kg_lease` / `prev_E` group singleton; add `window_pins`; per-group `cp_E`; `publish_ok`; steal at restore. #288/#290 as currently written (group lease, group `prev_E`, WRO `E ≤ serving_E AND NOT prev extras`) must not land.

anovv:
## Holes / blockers / fencing

Two correctness bugs, then fencing. The overlay/pin split is sound for **one** failover; it is not sound for republish or a second checkpoint.

### Blockers

1. **WRO republish cuts history.** `attempt == serving.attempt ∧ E ≤ serving_E` is right for idle keys (pin stays). It is wrong for a key in `dirty`. Source replay is post-CP only; `commit_events` does not rewrite the window; `> cp_wm` opens on the next watermark, not when catch-up is complete. First publish can be `B@150` on a key whose retained raw/tiles are still `A`. WRO then rebuilds without A. That is a wrong answer. Group `prev_E` was hole-fill for this. Per-key pins replace it only for untouched keys. You still need per-key `prev_*` on publish, or copy-on-first-touch before a key may go dirty. GC of “not overlay ∪ serving” then deletes the A rows you just dropped.

2. **Streaming idle keys die on the second checkpoint.** Overlay is one generation: `me | (cp_g.attempt ∧ E ≤ cp_E_g)`. A → B: B still sees A’s idle keys. B checkpoints `attempt=B` (even if `cp_E[g]` is inherited). C restores: overlay is C | B. A’s rows are invisible, source will not replay them (offset is B’s CP). Keyed state is gone. That is not a documented taste; it is failed recovery. You already store `O(groups)`. Persist per-group `(attempt, E)` and inherit A’s cut for untouched groups.

### Concurrency / fencing

What works: one WO per group; LWT on `owner` for steal/publish; WRO pins once per request (no mix mid-lookup); successor overlay hides `old_attempt ∧ E > cp_E`; after steal, zombie publish CAS-fails and kills that task; two restorers, one steal wins.

Holes:

- **Steal `IF owner = previous` must reread.** Previous is not in the CP blob. If B1 stole and died, owner is B1, serving still A. B2 CASing `IF owner = A` misses and you fail the task forever. Read owner; if not me, CAS from observed; miss → reread; if another live `me`, then fence.

- **Streaming zombies are unfenced in the store.** No `owner` row, no ingest LWT. Overlay only protects the **successor’s reads**. The dead worker still writes unlogged rows and still emits downstream until master kill. Request-mode publish CAS is only a backup, and only if cadence actually runs (slow `interval` → long write storm). Output fencing is job `attempt` on the data plane, not `window_pins`.

- **Grey window is real.** Until steal is applied, old owner can still publish. WRO can jump `A@100 → A@5001` after B has started. You accept that; do not also start ingest before steal finishes, or you widen it.

- **Unlogged zombie extras after steal.** A can still `commit_events`. Hidden from B/WRO if pins/overlay are right; they still bloat and can race GC. Not a read bug.

- **`publish_ok` vs replay is not a fence.** `current_wm > cp_wm` is one frontier tick. Replayed events can produce that tick on batch one. That is how (1) happens under concurrency, not a separate bug.

- **Chunked LWT.** One partition BATCH of unbounded `dirty` will not stay one Paxos. Partial publish is OK for WRO (per-key pin) **only if** you do not require `serving_wm` and all keys to move together. Spec that.

- **Same-partition `window_pins`.** `((ns, key_group), business_key)` is the unbounded group partition you refused for data. Steal/publish contend on one Paxos key with every key in the group. That is a liveness/hot-partition problem under steal-at-restore (`O(groups)` LWTs) and large `dirty`.

### Not blockers

- Seeding `E` from `cp_E` not `serving_E` (`B@150` vs `A@5000`) is fine if WRO never `max(E)` across attempts.
- `>= serving_wm` **and** `> cp_wm` is the right restore table.
- WRO pin-once vs concurrent publish: coherent old or new, not a mix on one key.
- Rescale disjoint groups: two new tasks cannot both own a group. Steal per group with **that** row’s owner.
- Stalled wm (`current_wm == cp_wm` forever): pins do not move; stale but consistent.

### Bottom line

Successor reads are fenced; zombie **writers** are not, except request publish CAS + master kill. The two must-fix correctness bugs are republish-without-holes and one-generation overlay wiping idle streaming state on the second CP. Everything else is fencing/ops around steal-read-CAS, grey window, and pin-partition Paxos.

anovv:
Superseded by #299, which keeps this PR's per-group cuts, rescale-as-slice-remap, three-clock restore, and `publish_ok` table, and replaces the opaque-`attempt` equality overlay with a totally ordered `(generation, seq)` version plus a per-key-group cut history. That closes both blockers found in review here (second-failover history loss, republish truncation), and lets serving state go back to one `window_pins` row per group with a single-row publish LWT — no per-key pins, no dirty set, no chunked Paxos. #299 also fixes the GC rule, which could never reclaim a superseded version under `attempt == me`. Leaving this open until #299 is reviewed.

anovv:
Superseded by #157, which is now the plan of record.

This PR's protocol matched `attempt` by equality, so the visible set was a union of per-attempt prefixes rather than a prefix of one order. That loses every cell an attempt wrote and its successor did not rewrite, on the second failover, and coverage planning reads a missing tile as an empty interval — an undercount rather than an error. #157 fixes it by making the version totally ordered and the cut a history.

What survived into #157: per-group cuts, rescale as slice remap with no data copy, the three-clock restore rule, the skinny `window_kg_buckets` GC index, the refusal to cluster `business_key` under a key group on data tables, and `LOCAL_QUORUM` data with LWT only where ownership is at stake.

PR #299, opened 2026-09-18
Document the cut-history Scylla window store protocol

anovv:
## Summary

New design PR for the Scylla window store. **Supersedes the protocols in both [#157](https://github.com/volga-project/volga/pull/157) and [#298](https://github.com/volga-project/volga/pull/298).** Spec only.

#298 stamped rows with an **opaque** `attempt` plus a per-group epoch and matched `attempt` by **equality**, so the visible set was a union of per-attempt prefixes rather than a prefix of one order. Review of that PR found two silent wrong-answer bugs, and both have the same root cause:

1. **History loss after the second failover.** `A -> B -> checkpoint -> C` leaves `C` with `C | (B AND E <= cp_E_B)`. Every cell `A` wrote that `B` did not rewrite is invisible, and replay only covers post-checkpoint input. The unit of loss is a **cell**, not an idle key, so active keys lose retained raw history too — and a missing tile is read as *an interval with no rows*, so the failure is an undercount rather than an error.
2. **Republish truncates a key's history.** With per-key pins and `attempt == serving_attempt`, the first publish touching a key hid every row an older attempt wrote for it. `prev_E` existed for exactly this; per-key pins only fixed the *idle*-key half.

## Protocol change

- `Version = (generation, seq)`, **totally ordered**. `generation` is durably monotonic and never reused; `seq` is a plain per-`key_group` counter from 0 (all epoch-seeding rules deleted).
- Visibility is a **per-key-group cut history**: a sorted list of one `(generation, seq)` prefix per generation. Old generations keep their prefix forever, so history survives arbitrarily many failovers; zombie writes stay above their generation's prefix and are excluded **permanently**. No `prev_E`, no copy-on-restore, no catch-up freeze.
- Serving state returns to **per key group**: `window_pins` becomes **one row per group, no clustering column**, and publish becomes a **single-row LWT**. That removes the unbounded dirty-key set, the chunked conditional BATCH, and the mega-partition every request point-got.
- `publish_ok` (`current_wm >= serving_wm AND current_wm > cp_wm AND cut advanced`) is unchanged and now **sufficient**, because it guards a complete history instead of a truncated one.
- **Clamp invariant**, written down explicitly: a successor's first publish clamps the predecessor's entry down to the checkpoint prefix, so the serving cut is monotonic in *event-time coverage*, not in row membership.

## Also fixed in this rewrite

- **Per-cell version retention in GC.** #298's "drop rows in neither overlay nor pin" could never reclaim a superseded version, since `attempt == me` matched every epoch of the current writer. Tiles are rewritten on every ingest batch that touches them and `key_state` on every batch, so `window_tiles` / `window_key_states` grew with write count on the hot read path. New rule: at most four versions per cell.
- **Version-level commit atomicity.** One commit spans 4–6 partitions and cross-partition batches are forbidden, so `commit_events` is not CQL-atomic. The acked-prefix rule is what hides torn commits; a timed-out write is retried at the **same** version and blocks the cut.
- **Newest-first version clustering** with bounded `LIMIT` reads for `key_state` (one query in steady state), available only because versions are totally ordered.
- Lexicographic `(generation, seq)` cell collisions, replacing `max(E)` then `max(attempt)`.
- Read-then-CAS steal (`previous` is not in the checkpoint blob), bootstrap rules, pin GC, and the `VersionedRange` type — all review items from #298.
- **Durable generation allocator** called out as a master prerequisite: `attempt_id` resets on master start, and `max(cut) + 1` repeats when two attempts restore the same checkpoint.

Retained from #298: per-group cuts, rescale as slice remap with no data copy, three-clock restore with `min(cp_wm)`, publish off the CP barrier, `window_kg_buckets` + `floor_bucket`, no `business_key` clustering on data tables, `LOCAL_QUORUM` data with LWT only for ownership.

Normative cut: *Protocol (normative)* in [`STORE_DESIGN.md`](https://github.com/volga-project/volga/blob/docs/window-store-cuts/src/runtime/operators/window/STORE_DESIGN.md). Stack, testing, and the ten protocol cases that need explicit tests: [`docs/window-store-impl.md`](https://github.com/volga-project/volga/blob/docs/window-store-cuts/docs/window-store-impl.md).

## Test plan

- [x] Documentation-only change
- [ ] Close #298 (protocol dead) and #157
- [ ] Land PR 0b (durable generation) — gates every versioned write
- [ ] Restack #287–#290 after merge: `(generation, seq)` descending row keys, single-row `window_pins`, cut histories, per-cell version retention. Both prior cuts are dead; the data-table row key changes, so restacking earlier means rewriting the schema twice.


anovv:
## Review

This is the right protocol. Totally ordered `(generation, seq)` plus a per-group cut *history* actually closes the two #298 silent wrong answers (second-failover cell loss, republish truncation). Single-row `window_pins`, read-then-CAS steal, acked-prefix, durable generation, and “no dirty set” are the correct fallout. Do not restack onto #298.

A few things will still produce wrong answers or unimplementable GC if an implementer follows the text literally.

---

### Blocking: publish must *replace* the serving cut from `cp_cut`, not `live.advance`

Clamp is the whole point of republish:

> B inherits `(A, cp_seq_A)` from the blob; live serving is `(A, serving_seq_A)`; first publish clamps `A` down to `cp_seq_A`.

`CutHistory::advance` is specified to **never raise (and never lower) older generations**. Then the LWT is:

```text
SET serving_cut = serving_cut.advance((my_generation, cut_top[g]))
```

That keeps `{A:5000}` and appends `B`. Clamp never happens. WRO keeps A's post-checkpoint rows, now alongside B's replay of the same input, with different `Cursor`s — duplicate raw, undercount/overcount depending on LWW.

Normative publish payload:

```text
serving_cut := cp_cut[g].advance((my_generation, cut_top[g]))
SET serving_cut, serving_wm IF owner = me
```

`cp_cut` already has `A` at the checkpoint prefix. Advance only adds/raises **me**. Ahead-of-CP serve clamps down; behind-of-CP serve fills A's prefix up to the blob (coverage does not go backwards). Later publishes of mine reuse the same formula (`cp_cut` is stable until the next barrier).

Write `advance` in full in the doc. It is the load-bearing function and currently `/* ... */`. Also: do not `advance((me, 0))` on a group with no acked write — that would insert a fake prefix.

Until this sentence is unambiguous, #290 will ship the #298 bug again.

---

### Blocking: GC grace (rule 4) is not implementable from stored state

Keep “newest allowed by any `serving_cut` published within grace.” The pin row **overwrites** `serving_cut` on every publish. After clamp, `{A:5000}` is gone. OnCommit can publish many times inside one WRO budget, so even `prev_serving_cut` only saves the last publish.

In-flight WRO that pinned `{A:5000}` then needs A's `(100, 5000]` until the request ends. GC that only sees current `{A:100, B:n}` will drop them.

Pick one and spec the extra state:

- On clamp, delay dropping that generation's `seq > new_seq` until `now > clamp_time + WRO_timeout` (no need to store old cuts; do not apply “above-cut” deletes for that generation during grace), or
- Keep a small ring of serving cuts / per-generation `max_seq_served` with a timestamp, rotated only after grace.

“Four versions per cell” is the right *bound*. Rule 4 as written is not a procedure.

---

### Writer-visible torn commits

Acked-prefix correctly hides torn `(gen, seq)` from **cuts**. WO visibility is still `row.generation == my_generation` with **no seq cap**, so the writer can `load_key_state` / eval / emit from a partial commit of `seq` still in flight.

“Complete or fail before checkpoint or publish” is not enough — due emit can happen first.

Either: even for me, readers use `seq <= acked_prefix[g]` except the in-flight writer of that `seq` retrying; or: do not ingest/advance further in that group until the in-flight `seq` acks or the task fails. Retry at the **same** version stays.

---

### Concurrency / fencing (remaining)

Sound: one owner LWT; steal reread; WRO pins the cut once; successor never raises an old generation; zombie `seq` above that generation's entry is excluded forever; two restorers, one steal wins; `Not applied` / LWT timeout fails the task.

Still true (you already listed 4–5; keep them):

- Steal before ingest, or the grey window widens.
- Streaming zombies are unfenced as **writers** (unlogged extras + downstream) until master kill. The cut only fences readers.
- Publish LWT timeout must be treated as unknown (same as steal): fail, do not retry a second `seq` as if it were a new publish.
- Dense `seq` per group: a hole (`5` timed out, `6` acked) means `cut_top` cannot skip `5`. Allocate the next `seq` only after the previous is acked, or the low-water mark is the protocol. HOL blocking of the whole group's publish is accepted; skipping is not.

Durable generation is a real gate. `attempt_id` and `max(cut)+1` are both wrong as you wrote. Backend must refuse to open without it. Persist **before** `Configure`.

---

### Spec nits (fix here)

- `advance` body, and the publish assignment above, belong in *Protocol (normative)*.
- Checkpoint `cuts[g].advance((me, cut_top))` for a group with **no** acked write: leave inherited entries, do not insert `(me, 0)`.
- `key_state` walk is `1 + |cut|` queries on an idle key, not “two just after restore.” After 100 failovers it is 100. The cap-and-fail on history length is then a **read** cap, not just blob size. Say that next to the CQL.
- Grace vs WRO pin-cache TTL: good. Tie it to the implementable rule once you pick one.
- `VersionedSlice` markdown is still truncated (`Vec `).
- Bootstrap `publish_ok` skipping both wm compares: good, keep.

---

### What not to reopen

- Per-key pins / dirty sets / chunked LWT BATCH.
- Seeding `seq` from `cp` or serving.
- `max(E)` then `max(attempt)`.
- Any-generation `seq <= cp_seq` overlay.
- Publish on the CP barrier.

Per-cell retention (newest me / cp / serving / grace) is the right replacement for “in the overlay or pin.” Tests 1–4, 5–6 (torn / timeout block the cut), 8 (generation monotonicity) in the impl doc are the ones that matter; 1–2 are the #298 bugs and should fail a #298 implementation.

Restack #287+ only after publish := `cp_cut.advance(me)` and GC grace are written as procedures, and PR 0b exists. Schema change is real (`(generation, seq)` DESC); do not rewrite it twice.

anovv:
Agree with the review above on clamp (`serving := cp_cut.advance(me, cut_top)`, not `live.advance`), GC grace as a procedure, writer torn `seq` cap, untouched groups, and `key_state` walk = `1 + |cut|`. Not repeating those.

Leftover that still needs a sentence in *Protocol (normative)*:

1. **`advance` a copy.** `serving_cut := cp_cut[g].advance((me, cut_top))` must not mutate the in-memory `cp_cut` used for WO overlay. If `advance` is `&mut self`, clone first. Otherwise a publish quietly turns the restored checkpoint into the live serve prefix.

2. **`next_seq[g]` under concurrent `commit_events`.** Ingest can run `process_key_concurrency` keys in one group. `seq = next_seq[g]++` needs an atomic allocator. Pipelining is allowed only with the dense acked-prefix already specified; two in-flight seqs in the same group without that is a torn `load_key_state` on a *different* key sharing the group clock.

3. **Pin payload cap.** `serving_cut` on the single-row LWT grows with the history (`O(generations)` × 16 B), same as the blob and the `key_state` walk. The cap-and-fail is also a **publish** cap, not only a read/blob cap.

4. **Markdown.** Same angle-bracket eat as `VersionedSlice`: recovery still says `Vec `; `docs/window-store-impl.md` steal is `IF owner = ` with the observed owner dropped.

No disagreement on restack: wait for publish := `cp_cut.advance(me)`, implementable GC grace, and PR 0b.

anovv:
Both reviews are right on substance. Pushed 61eecccc addressing everything, with two corrections where the diagnosis holds but the prescribed fix would introduce a different bug.

## Fixed as specified

**Publish must base on `cp_cut`, not the live cut.** Correct, and it was a self-contradiction: the prose said B's first publish clamps A down, the CQL said `serving_cut.advance(...)`, and `advance` never touches another generation's entry — so the clamp provably never fired. Worse than #298's truncation, because serving would then hold `{A:5000, B:n}` and WRO would return A's post-checkpoint rows *next to* B's replay of the same input with different `seq_no` cursors: duplicate raw, so an overcount.

Now normative:

```text
next = cp_cut[g].advance(my_generation, cut_top[g])
```

`advance` has a full body, takes `&self` and returns a new history (which also closes the "publish mutates `cp_cut`" hole from the second comment — same bug from the other side), and no-ops on `cut_top == None` rather than writing a `(me, 0)` entry, since `seq` is 0-based. There's a table for the three restore cases (ahead clamps down, behind fills up to the acked blob prefix, later publishes raise only my own entry).

One case neither comment raised, now covered: a generation can be **dropped** rather than lowered. One that stole and published but died before any checkpoint never enters any `cp_cut`, so the next publish omits it entirely. Correct for the same reason, and it's why the grace test is "lowers **or** drops".

**GC grace needed a procedure.** Correct — the pin row is overwritten on every publish, so "any `serving_cut` published within grace" exists nowhere GC can read. Went with option (b), but with the fix that makes a single slot sufficient: `prev_serving_cut` + `prev_expires_at` on the pin row, written **only** by a publish where `observed_serving_cut.outlives(next)`.

That answers the objection to a depth-1 ring. Ordinary publishes only raise the publisher's own entry, so their predecessor is a strict subset of the new cut and retaining it is a no-op; writing `prev` unconditionally is what breaks, because a clamp followed by two ordinary publishes pushes the clamped cut out of the slot while a reader is still pinned to it. Coverage-removing publishes happen once per failover, so one slot plus an expiry covers every ordering. It lives in the pin row rather than owner RAM so a successor's GC honours a grace window opened by its predecessor. Grace now explicitly includes a clock-skew allowance, since `prev_expires_at` is absolute wall clock written by one node and read by another.

**Nits, all in.** No `(me, 0)` for untouched groups (checkpoint path too). Atomic per-group `seq` allocator. Publish LWT timeout is unknown — fail, and explicitly do not retry as a fresh publish at a later prefix. Dense prefix with the worked hole example (`5` times out, `6` acks, `cut_top` is `4`, never skip, never allocate around it) and HOL blocking of that group's cut folded into the accepted-divergences entry. History cap now states its four dimensions: blob, `serving_cut` payload on the LWT, per-row filter cost, and the pathological `key_state` fallback.

## Diagnosis right, fix corrected

**Writer-visible torn commits.** Real gap — requiring resolution before checkpoint or publish doesn't cover the advance/emit read. But capping the writer at `seq <= cut_top[g]` breaks read-your-writes: with X's `seq=5` unresolved and Y's `seq=6` acked the prefix is `4`, so Y's own committed state goes invisible to its writer, which re-ingests and re-applies it. Quiescing the group is safe but forbids the cross-key concurrency `ingest_key_concurrency` exists for, and with `max_p` groups concurrent keys collide often.

The invariant only needs to be **per key**, since a torn commit is only observable by mixing partitions of the same key:

> While a commit for key `K` is in flight, the writer issues no read of `K` and no second commit for `K`. `commit_events` either returns with every write acked, or the task fails.

Cross-key concurrency inside a group stays legal; the low-water prefix already handles the rest, and the doc now explicitly rejects the capped-writer variant with the reason.

**`key_state` walk.** The count is right and "two just after restore" was wrong — 101 queries after 100 failovers. But the remedy isn't to accept the walk and call the history cap a read cap; it's to delete the walk. Clustering is already `(generation DESC, seq DESC)` and per-cell retention bounds the partition to four versions, so one `LIMIT 8` read with a client-side filter resolves it in a single query. The per-generation restricted query survives only as a fallback (double the limit to a cap, then walk) for a key a zombie rewrote many times pre-sweep. So the cap bounds blob, payload, and filter cost — not steady-state query count.

## Not a real defect

The `Vec<...>` / `IF owner = <observed>` truncations aren't in the files: all three were inline code spans and `VersionedSlice` is inside a ```rust fence, so they render correctly on GitHub. Both reviews are reading a scraped conversion of the page that eats angle brackets — the same mangling visible in the #298 dump. Reworded anyway so it stops getting re-reported.

## Tests and restack

Added to the impl doc: test 4 now asserts the published cut equals `cp_cut.advance(B, cut_top)` and **not** `live_serving.advance(...)`, including a generation whose entry is dropped rather than lowered; new test 4b covers grace after a clamp (pin, clamp, `maintain`, finish the read, then expire) plus "an ordinary publish must not overwrite the grace slot"; test 7 adds the idle-key-after-many-generations case that catches a regression to the per-generation walk.

Agreed on the gate: #287+ restacks only after PR 0b exists. Both blockers are now written as procedures rather than prose.

anovv:
## Review of `61eecccc`

The previous blockers are closed in the text: `next = cp_cut.advance(me, cut_top)`, `advance` is non-mutating and never inserts `(me, 0)`, torn commits are serialized **per key**, prefix does not skip holes, `prev_serving_cut` is written only on `outlives` / coverage removal. That is the protocol to implement.

Two leftovers will still mis-implement fencing/grace if copied literally.

### 1. Observe serving **after** steal applies, not from `SELECT owner`

Steal is:

```text
SELECT owner ...
UPDATE SET owner = me IF owner = row.owner
```

Publish then says observed serving came from “the row at steal.” That SELECT does not read `serving_cut` / `serving_wm`, and if it did **before** the CAS, the grey window can still raise serving (`A:5000 → A:6000`) before owner moves. `prev_serving_cut` would then be the stale `5000` while in-flight WRO pinned `6000`; GC grace misses the rows.

After steal **applies**, A can no longer publish, so serving is stable until B’s own LWT. Normative:

```text
steal CAS (owner only)
then read the full pin row   -- serving_wm, serving_cut, prev_*
observed_serving_cut := that serving_cut
```

Or fold the read into the LWT result. Do not use the pre-steal SELECT as the grace snapshot. `publish_ok` also needs post-steal `serving_wm`.

### 2. `published_top` is `Option`, like `cut_top`

`seq` is 0-based. After the first acked write, `cut_top = Some(0)`. If `published_top` is initialized to `0`, then `0 > 0` and the first publish never fires (bootstrap and post-restore). Init `published_top = None`; `Some(0) > None`. Same class of bug as not inserting `(me, 0)`.

---

### Closed (do not reopen)

- Clamp via `cp_cut.advance`, not `live.advance`. The three-row table is the implementer test.
- `advance(&self, me, Option<seq>)` + checkpoint `None` leaves inherited entries.
- Per-key in-flight: no read/second commit of `K`; do **not** cap the writer at `cut_top`.
- Dense prefix / HOL on timeout.
- Grace: one `prev_*` slot, only when `observed.outlives(next)`; ordinary raise-me publishes must not rotate it. Clock-skew in `prev_expires_at` is accepted.
- `key_state` `LIMIT 8` then fallback; no per-generation walk on the hot path.

### Still accepted (keep in the doc)

Grey window until steal, streaming writer zombies, durable generation as PR 0b, cut-history cap, non-deterministic `seq_no`. Steal-before-ingest stays mandatory.

Nit: impl-doc pin summary still lists `owner, serving_wm, serving_cut` without `prev_*`. Test 4b is the right grace test.

Ready to merge the design once (1) and (2) are one-line fixes in *Steal* / `publish_ok`. Then PR 0b, then restack with this schema — not #298’s.

anovv:
Follow-up on `61eecccc`. Clamp, `&self` advance, `(me, 0)`, per-key torn rule, atomic `seq`, publish-timeout-as-unknown, and `prev` only on `removes_coverage` all look right. Not reopening those.

### Blocking: `load_key_state LIMIT 8` is not a correctness bound

```cql
SELECT generation, seq, key_state FROM window_key_states
 WHERE namespace = ? AND key_group = ? AND business_key = ?
 LIMIT 8
Keep the first returned row that passes the filter.
```

That is only safe **after** per-cell retention has already left ≤4 versions. Retention is periodic. Until then the partition is one row per ingest.

Idle key, `A` wrote seq `1..20`, CP at `10`, post-CP `11..20`, no GC yet, `cp_cut = {A:10}`:

- clustering is `(generation DESC, seq DESC)`
- `LIMIT 8` returns `A@20..13`
- none of those is `me`, none has `seq ≤ 10`
- result: empty `KeyState` → new seq line → **lost window** (same class of silent undercount as #298)

`LIMIT 8` is a perf hope, not a filter. Either page until a visible row or end of partition, or keep a targeted `generation = ? AND seq <= ? LIMIT 1` for `me` and each `cp_cut` entry, or compact that partition before the read. Do not ship “one query, LIMIT 8” as the normative hot path.

### Leftover (smaller)

1. **Steal `SELECT` vs `observed_serving_cut`.** Steal is written as `SELECT owner`. Publish needs `serving_cut` for `outlives` / `prev`. Steal must read the whole pin row (or publish re-reads it). “No extra read” is false if steal only fetched `owner`.

2. **Two coverage-removing publishes in one WRO.** One `prev` slot: `{A:5000}` → B clamp (prev=`{A:5000}`) → B dies pre-CP → C publish drops B and **overwrites** prev with `{A:100, B:n}`. In-flight WRO still pinned to `{A:5000}` loses grace. Rare (two failovers inside one request). Safer: if `prev` is unexpired **and** `prev.outlives(observed)`, keep `prev`.

3. **RAM table** should list in-flight **keys** (bounded by `ingest_key_concurrency`), not only in-flight seqs.

4. **`commit_events` acked-or-fail vs retry-same-Version.** Fine if retry is **inside** the call and the per-key lock is held until success/fail. Say that so nobody returns on timeout and then `load_key_state`.

Restack still waits on the `LIMIT 8` fix. Everything else from the previous comments is closed.

anovv:
Follow-up on the `LIMIT 8` comment (https://github.com/volga-project/volga/pull/299#issuecomment-5729254119). Overlap with the previous `61eecccc` note is small; they are mostly complementary. One blocking claim is overstated.

**Overlap.** Steal is `SELECT owner` but publish needs `serving_cut` for `outlives` / `prev`. Both right that “no extra read” is false. That comment is completeness (`outlives` has no input). The stronger hole is **timing**: even `SELECT *` *before* the CAS can be stale, because A can still publish in the grey window (`5000 → 6000`). Observe serving **after steal applies** (LWT result, or a full-row read after the CAS). Pre-steal `SELECT *` is not enough. `publish_ok` needs that post-steal `serving_wm` too.

**`LIMIT 8` — diagnosis yes, empty `KeyState` / restack gate no.** Idle key, A wrote `1..20`, CP at `10`, no GC: `LIMIT 8` returns `A@20..13`, none visible. True of the first query *alone*. The next paragraph already falls back: double the limit, then `generation = ? AND seq <= ? LIMIT 1` per history entry. That finds `A@10`. Empty `KeyState` only if someone copies the first CQL block and skips the fallback.

Do not ship `LIMIT 8` as the *only* read; do not treat the fallback as optional. Do not hold restack on rewriting the hot path. The copy-paste hazard is real; silent undercount is not, if the fallback is implemented.

**That comment, not in the previous note — all valid, none restack-blocking:**

1. Two coverage-removing publishes in one WRO overwrite `prev` (`{A:5000}` → B clamp → C drops B). One slot cannot keep two distinct in-flight pins. Keep `prev` if it is unexpired and `prev.outlives(observed)` saves the **oldest** pin, not a second WRO that pinned B’s cut. Write the keep-prev rule; a ring is optional.
2. RAM table should list in-flight **keys** (bounded by `ingest_key_concurrency`), not only seqs. The per-key mutex is in prose; the table currently omits it.
3. Retry the same `Version` **inside** `commit_events`, lock held until ack/fail, so a timeout cannot return and then `load_key_state`.

**Previous note, not in that comment:**

1. Post-steal observe (above), not only “read the whole row.”
2. `published_top` is `Option`, like `cut_top`. Seq is 0-based. After the first ack, `cut_top = Some(0)`. If `published_top` starts at `0`, then `0 > 0` and the first publish never fires (bootstrap and post-restore). Same class as `(me, 0)`. RAM table currently says scalars.
3. Impl-doc pin summary still omits `prev_*`. Nit.

Restack gate stays: post-steal observe, `published_top: Option`, optionally the keep-`prev` sentence, then PR 0b. Not the `LIMIT 8` rewrite.

anovv:
Superseded by #157, which is now the plan of record.

The cut-history protocol here is correct and carries over unchanged. What #157 drops is everything downstream of the separate publish path: the ownership lease, steal, `publish_ok`, the publish LWT, per-key pins, `prev_serving_cut` with its wall-clock grace, and the clamp invariant.

Tying publication to checkpoint completion removes the need for all of it. The published cut becomes exactly the cut a successor restores from, so it only ever grows and can never name writes that get replayed — which is the only reason the clamp existed. That trades some serving freshness for roughly half the protocol, and `ReadOptions::Fresh` buys the freshness back on the read path where it is one filter branch instead of a distributed protocol.

Renamed `generation` to `attempt` and kept `epoch`, matching the terms already in the codebase.

PR #302, opened 2026-09-21
Align request-mode checkpoints with reachable tasks (#301)

### LINKED ISSUE #301: Checkpoints never complete in request mode: read path is in the align set but gets no barrier
## Problem

In request mode a checkpoint can never complete. The read path is in the barrier-alignment set but can never receive a barrier, so every checkpoint times out and the master recovers in a loop.

- The master builds `expected_aligns` from **every** task in the graph (`master/state.rs:248-255`), while `expected_acks` is only the checkpointable ones (`:247`).
- Completion requires both sets to be fully satisfied (`master/checkpoint/protocol.rs:132`).
- Barriers are injected only into checkpointable **sources** (`worker/lifecycle.rs:102-106`). `HttpRequestSourceConfig`, `WindowRequestOperator` and `RequestSinkConfig` are all `false` in `operator_config_requires_checkpoint` (`operators/operator.rs:132-144`).
- The read path is a **disjoint component**: `to_request_mode` removes the WO's outgoing edge and re-attaches the followers to the WRO (`api/logical_graph.rs:480-483`). No barrier from a streaming source can reach it.

So HTTP source, KeyBy, WRO, projection and request sink sit in `expected_aligns` forever → timeout → `Recover(HashSet::new())` (`master/attempt/execute.rs:206-207`) → new attempt → repeat.

Masked today only because `CheckpointSpec::default()` sets `interval_ms: Some(0)` (`api/spec/state.rs:54-62`) and no request-mode test or benchmark enables an interval.

## Why now

The Scylla window store publishes the serving cut **on checkpoint completion** ([#291](https://github.com/volga-project/volga/issues/291), [#300](https://github.com/volga-project/volga/issues/300)). Request mode therefore requires a non-zero checkpoint interval, which is exactly the configuration that triggers this. Without the fix the backend cannot publish at all.

## Plan

Align only what a barrier can reach. Candidates, cheapest first:

- Compute `expected_aligns` as the tasks reachable from checkpointable sources, instead of `all_tasks()`.
- Or exclude request-subgraph vertices explicitly; they hold no state and are already excluded from `expected_acks`.

Either works today; after [#247](https://github.com/volga-project/volga/issues/247) splits the graph it becomes per-subgraph and the reachability walk is trivial.

Test: a request-mode pipeline with a non-zero `interval_ms` must complete checkpoints and keep serving. Nothing covers this combination now.

## Blocks

Request mode on the Scylla backend ([#291](https://github.com/volga-project/volga/issues/291)).

anovv:
## Summary
- `expected_aligns` is now the downstream closure of checkpointable sources instead of `all_tasks()`.
- Request-mode HTTP source / WRO / request sink sit in a disjoint component and are no longer in the barrier-alignment set, so a non-zero checkpoint interval can complete.
- Unit tests cover streaming (aligns = whole graph) and request mode (read path excluded).

Fixes #301. Independent of #247 worker split; that stack bases on this.

## Test plan
- [ ] `request_mode_align_set_excludes_read_path` and `streaming_align_set_is_the_whole_graph`
- [ ] Existing local checkpoint suite (interval checkpoints still complete on streaming graphs)


anovv:
Closing this. The reachability walk is not how #301 gets fixed.

#303 takes the read path off the streaming graph (keyby, window request, and the SQL followers become a `RequestChain`). `expected_aligns` stays `all_tasks()` of that streaming graph, which is exactly the set a barrier can reach. #304–#306 are rebased onto master through #303.

PR #303, opened 2026-09-21
Split request compile into two graphs (#247 PR1)

anovv:
## Summary
- `to_request_mode` now extracts a request `LogicalGraph` (operator parallelism 1) and leaves the streaming write path on `self`.
- `compile_pipeline` returns `{ streaming, request }`; `compile_logical_graph` is the streaming graph (what the master uses).
- In-process `RequestExecutor` runs the request chain as one tokio task per HTTP request (semaphore + worker). HTTP source/sink `StreamTask`s are not used on this path.
- Planner / compiler tests assert the split and `state_owner_operator_id`. Request source e2e uses `RequestExecutor`.

Stacked on #302. Plan: `docs/issue-247-request-workers.md`.

## Test plan
- [ ] `request_compile_emits_two_graphs`, `test_window_query_to_request_mode`
- [ ] `test_request_source_sink_e2e`
- [ ] `request_mode_align_set_is_streaming_graph_only` (from #301, updated for the split)


anovv:
Pushed `e2904423`: the HTTP source/sink is no longer an operator or a connector.

`PipelineSpec.request` is a `RequestSpec` (`max_pending_requests`, `request_timeout_ms`). Compile returns a `RequestChain` (keyby → window request → SQL followers, plus the window input schema). Listen address stays per replica. Deleted `RequestSourceSinkSpec`, the request source/sink functions, `RequestRoute`, and the worker request-source processor.

#304–#306 are rebased onto this. The request worker starts that chain; metrics and the kube/docker replica sets are unchanged in intent.

